PR #231, opened 2024-01-16
[ENG-580] build: add ApiService layer to combine generated services

dionlow:
### build: add ApiService layer to combine generated services
run `yarn generate-api`, updates api service layer with api refactor. 
 
#### problem
The separated services (`AllowedDomainApi`, `Configuration`, `ConnectionApi` ...) no longer support a single defaultApi endpoint. 

#### Trade offs
**Separated Services:** As the service layer becomes more complex on the backend, modular separation of services is easier to manage and document, but comes at the cost of complexity the client developer and codebase must understand. 

#### Solution
We need to create a service layer to begin managing the service level changes. This PR adds a ApiService class to combine the service endpoints. Note all current endpoints are added into the class, even services not currently used. This file could be reduced down or copied over to the client.

#### Future
It may be worth adding a custom script to generate this file after 'yarn generate-api' is run.
`src/services/ApiService.ts`

#### Other trade offs

**Dynamic Combined Service** 
This potential service layer would allow us to call  `apiService.listInstallations()` instead of `apiService().installationService.listInstallations()`.This support would reduce the client's need to understand backend logic, however **we choose to avoid this for now**, as it may add additional complexity and difficulty while debugging and maintaining code since this step is not done by a tested and supported library of converting generated services and types into a combined service. 

It is possible to create a dynamic service layer with the following for future reference
``` 
function createDynamicApiService(config: Configuration) {
  const service = new ApiService(config);
  return new Proxy(service, {
    get(target: ApiService, propKey: keyof ApiService, receiver) {
      const value = target[propKey];

      // If the value is a function, bind it to target and return
      if (typeof value === 'function') {
        return value.bind(target);
      }

      // If value is an object, check for the method and bind it
      if (typeof value === 'object' && value !== null) {
        // Explicitly cast value to any to bypass TypeScript's strict type checking
        const method = (value as any)[propKey];
        if (typeof method === 'function') {
          return method.bind(value);
        }
      }

      // If none of the above, return the property as is
      return value;
    },
  });
}
```


src/services/api.ts
```diff
@@ -75,10 +77,10 @@ const config = new Configuration({
   basePath: AMP_API_ROOT,
 });
 
-let apiValue = new DefaultApi(config);
+let apiValue = new ApiService(config);
 
 // For testing, etc. we may want to use a different API configuration than the default
-export const setApi = (api: DefaultApi) => {
+export const setApi = (api: ApiService) => {
```
dionlow, line 83: This may or may not break tests if they are written with new service layer. @laurenzlong

laurenzlong (approved):

PR #233, opened 2024-01-17
[ENG-512]refactor: read state

dionlow:
### preps for write state to be separate section
- separates read state into own field
- preps for write state field

app should work as expected

#### steps
1. ~UI write tab + form~
2. **refactor read state**
3. add write state
4. connect data to write state slice
5. submit write state slice

### DIFF
--- a/src/components/Configure/fields/OptionalFields/setOptionalField.ts
+++ b/src/components/Configure/fields/OptionalFields/setOptionalField.ts
@@ -8,17 +8,22 @@ function setOptionalFieldProducer(
   fieldKey: string,
   checked: boolean,
 ) {
-  const draftSelectedOptionalFields = draft?.selectedOptionalFields || {};
+  const draftSelectedOptionalFields = draft?.read?.selectedOptionalFields || {};
   draftSelectedOptionalFields[fieldKey] = checked;
   if (!checked) {
     delete draftSelectedOptionalFields[fieldKey];
   }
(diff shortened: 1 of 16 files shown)


src/components/Configure/fields/OptionalFields/setOptionalField.ts
```diff
@@ -8,17 +8,22 @@ function setOptionalFieldProducer(
...
+
+  if (draft.read?.savedConfig.optionalFields) {
+    const savedFields = draft.read.savedConfig.optionalFields;
+    const updatedFields = draftSelectedOptionalFields;
+    const isModified = !isFieldObjectEqual(savedFields, updatedFields);
+    // immer exception if we try to set a value
+    // eslint-disable-next-line no-param-reassign
+    draft.read.isOptionalFieldsModified = isModified;
+  } else {
+    console.warn('read.savedConfig.optionalFields is undefined');
```
laurenzlong, line 25: is this going to always log for integrations that don't have optional fields? if so, that might be kind of annoying.

laurenzlong (approved):

src/components/Configure/fields/OptionalFields/setOptionalField.ts
dionlow: This code should not be reachable. I put this here for debugging purposes, if this becomes a case that I need to look at in the future.

PR #238, opened 2024-01-18
[ENG-512] feat: connect write form to state

dionlow:
### Summary
- removes dummy data
- connects write form state to context
- refactors field state to use a `useField` hook. 

#### demo with hydratedRevision data
write-field-state

#### steps
1. ~UI write tab + form~
2. ~refactor read state~
3. ~add write state to context~
4. **connect data to write state slice** (in progress)
4a. **connect hydratedRevision data**
4b. connect config data
5. submit write state slice
6. add isModified / in progress state

### DIFF
--- a/src/components/Configure/content/fields/useFields.tsx
+++ b/src/components/Configure/content/fields/useFields.tsx
@@ -1,0 +1,18 @@
+import { useProject } from '../../../../context/ProjectContext';
+import { useSelectedObjectName } from '../../ObjectManagementNav';
+import { useConfigureState } from '../../state/ConfigurationStateProvider';
+import { getConfigureState } from '../../state/utils';
+
+export const useFields = () => {
+  const { appName } = useProject();
+  const { objectConfigurationsState, setConfigureState } = useConfigureState();
+  const { selectedObjectName } = useSelectedObjectName();
+  const configureState = getConfigureState(selectedObjectName || '', objectConfigurationsState);
+
+  return {
+    appName,
+    configureState,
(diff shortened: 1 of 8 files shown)


src/components/Configure/content/fields/useFields.tsx
```diff
@@ -0,0 +1,18 @@
+import { useProject } from '../../../../context/ProjectContext';
+import { useSelectedObjectName } from '../../ObjectManagementNav';
+import { useConfigureState } from '../../state/ConfigurationStateProvider';
+import { getConfigureState } from '../../state/utils';
+
+export const useFields = () => {
```
laurenzlong, line 6: I wonder if `useConfigure` might be a more suitable function name?

laurenzlong (approved):
Woo hoo! We're getting there! Can you also add one more step to the task list to make sure the UI library still looks good when there are only write actions and no read actions. Just having a single "Other" tab might look weird. So you can either remove the LHN altogether, or rename "Other" to "Write" or something in that situation.

src/components/Configure/content/fields/useFields.tsx
dionlow: I think that makes sense, but I think it might get confusing with useConfigureState(). I might think of renaming both hooks. the current naming convention hopefully scopes it to just the fields folder for the time being.

dionlow:
added ticket: https://linear.app/ampersand/issue/ENG-587/feat-support-only-write-no-read

PR #253, opened 2024-01-31
[ENG-524] feat: adds redirectUrl support to Connections Provider

dionlow:
### summary
When redirectUrl param is present, the component will render a loading screen while redirecting to the URL. Otherwise, the child component will render.
- creates redirect handler component

#### test in mailmonkey
1. enter in a relative url (i.e. `my.redirect.url`)
2. go to `http://localhost:3001/connect`
3. fill out provider credentials
4. observe loading and redirect when credentials accepted

#### demo

redirect-handler

### DIFF
--- a/src/components/RedirectHandler/RedirectHandler.tsx
+++ b/src/components/RedirectHandler/RedirectHandler.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect } from 'react';
+import { Text } from '@chakra-ui/react';
 
 type RedirectHandlerProps = {
   redirectURL?: string;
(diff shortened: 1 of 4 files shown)


src/components/RedirectHandler/RedirectHandler.tsx
```diff
@@ -0,0 +1,31 @@
...
+  useEffect(() => {
+    // Check if a redirect URL is present
+    if (redirectURL) {
+      // Redirect to the specified URL
+      window.location.replace(redirectURL);
+    }
+  }, [redirectURL]);
+
+  // show a loading message if a redirect URL is present
+  if (redirectURL) return <Text textAlign="center">redirecting...</Text>;
```
laurenzlong, line 27: Can we use the LoadingIcon instead?

laurenzlong (approved):

PR #254, opened 2024-01-31
[ENG-524] feat: adds onSuccess connection handler

dionlow:
### summary
creates onSuccess and onError hooks based on whether a connection has been established and whether a boundary error is triggered. 
- adds onSuccess connection handler
- adds onError connection handler for ConnectionList error

error only works for ConnectionList error. onErrors don't support other errors like subdomain errors, closing dialogue too early, etc... expected experience needs more clarity or to be returned back to support more onError scenarios.

#### testing
- test by passing in an onSuccess in mailmonkey
- onError not fully tested yet

*NOTE: onError removed from PR*

### DIFF
--- a/src/components/Connect/ConnectProvider.tsx
+++ b/src/components/Connect/ConnectProvider.tsx
@@ -11,13 +11,12 @@ interface ConnectProviderProps {
   groupRef: string,
   groupName?: string,
   redirectUrl?: string,
-  onSuccess: (connectionID: string) => void;
-  onError: (error: string) => void;
+  onSuccess?: (connectionID: string) => void;
 }
 
 export function ConnectProvider(
   {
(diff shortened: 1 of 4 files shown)


src/components/Connect/ConnectProvider.tsx
```diff
@@ -36,11 +11,13 @@ interface ConnectProviderProps {
   groupRef: string,
   groupName?: string,
   redirectUrl?: string,
+  onSuccess: (connectionID: string) => void;
```
laurenzlong, line 14: Should be optional params

laurenzlong (approved):
Since right now we would only be able to detect ConnectionsList errors (and it sounds like it would be quite a bit of work to handle other error types), let's descope the `onError` handler for now. Having an error handler that isn't able to handle all errors is worse than not having one and will be confusing for builders. Good to merge once `onError` is removed.

dionlow:
removed onError code.

PR #267, opened 2024-02-07
[ENG-636] fix: replace loading with integration not found error message 

dionlow:
### Summary
The page will forever load if integration list is loaded but does not match an integration name. We now replace the loading icon with an error dialogue. 

Screenshot 2024-02-07 at 1 59 26 PM


jlimatampersand (approved):
LGTM

PR #272, opened 2024-02-16
chore: export needed contexts and clean up code

lino-levan:


src/hooks/useIsIntegrationInstalled.ts
```diff
@@ -5,10 +5,13 @@ import { useIntegrationList } from '../context/IntegrationListContextProvider';
...
-  isLoaded: boolean;
-  isIntegrationInstalled: boolean | null;
-}
+type UseIsIntegrationInstalledResult = {
+  isLoaded: false;
+  isIntegrationInstalled: null;
+} | {
+  isLoaded: true;
+  isIntegrationInstalled: boolean;
+};
```
lino-levan, line 14: The previous type was incorrect. There are two possible states for `UseIsIntegrationInstalledResult`, which is provided here. Checking `isLoaded` is enough to know whether `isIntegrationInstalled` is `null` or not.

src/components/Configure/nav/ObjectManagementNav/index.tsx
```diff
@@ -93,7 +92,7 @@ export function ObjectManagementNav({
           {isNavObjectsReady && (
             <Tabs
               index={tabIndex}
-              onChange={handleTabsChange}
+              onChange={setTabIndex}
```
lino-levan, line 95: I think this is a lot easier to sight-read.

src/hooks/useIsIntegrationInstalled.ts
dionlow: Possibly not incorrect, but the change looks more clear and precise!

dionlow (commented):
Changes look good, can you add a description of how these exported contexts are needed? 

Since the implementation may change in future; it's good to have context on these changes to look back on even if the changes are minimal. It would also be a good idea to link the linear ticket to tie these changes from multiple repo's together.

dionlow (approved):
nice, good work!

PR #276, opened 2024-02-20
refactor: move ConfigureInstallation to own folder

dionlow:
### Summary
We need to add some logic to control flow the UI to skip configuration step (when hydrated revision read and write is not present). We separate the "configuration step" into a refactored component named `ConfigureInstallation` which now contains 
```
  <ConfigurationProvider>
      <ObjectManagementNav>
        <InstallationContent />
      </ObjectManagementNav>
    </ConfigurationProvider>
```
and along with the `actions`, `content`, `nav`, and `state` folder which were previously one level up at the InstallIntegrations component level and 

This may affect builds relying on internal paths related to Configure Components.


dionlow:
refactor not needed at this time.

PR #277, opened 2024-02-20
fix: process undefined error

laurenzlong:
I was still getting a process undefined error, this is because when the `process` variable is undefined `(process?.env?.REACT_APP_AMP_SERVER)` still hits a runtime error. To see this in action, just open up developer console in the browser, and type in `process?.env?.REACT_APP_AMP_SERVER`, you'll get `Uncaught ReferenceError: process is not defined`. But `typeof process === 'undefined'` is safe to run.

### DIFF
--- a/src/services/api.ts
+++ b/src/services/api.ts
@@ -37,6 +37,12 @@ import { ApiService } from './ApiService';
 const VERSION = 'v1';
 
 function getApiEndpoint(): string {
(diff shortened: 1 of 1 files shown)


eberle1080 (approved):

lino-levan:
Ah, right. I should have used `globalThis`. This works too.

PR #279, opened 2024-02-21
[ENG-654] feat: add conditional proxy layout

dionlow:
### Summary 
We need control logic to allow `<InstallIntegration/>` to skip configuration step if an installation only has proxy actions. This control sequence should activate if no read and no write is present in the hydratedRevision, but proxy actions also occur.

This PR creates the layout based which future PR's will 
- [x] **add conditional layout render for proxy support** 
- [ ] add the `SuccessTextBox`
- [ ] add control logic for only proxy present
- [ ] add the createInstallation call

### DIFF
--- a/src/components/Configure/layout/ConditionalProxyLayout.tsx
+++ b/src/components/Configure/layout/ConditionalProxyLayout.tsx
@@ -1,0 +1,25 @@
+import { LoadingIcon } from '../../../assets/LoadingIcon';
+import { useHydratedRevision } from '../state/HydratedRevisionContext';
+
+interface ConditionalProxyLayoutProps {
+  children: React.ReactNode;
+}
+
+/**
+ * if the hydratedRevision does not have read or write,
+ * then it will not render the ConfigureInstallation
+ * @returns
+ */
+export function ConditionalProxyLayout({ children }: ConditionalProxyLayoutProps) {
+  const { hydratedRevision, loading } = useHydratedRevision();
+  const isProxy = !hydratedRevision?.content.read && hydratedRevision?.content.write;
+
+  if (loading) return <LoadingIcon />;
+  if (isProxy) return <div>Proxy Success</div>;
+
+  return (
+    <div>
+      {children}
+    </div>
+  );
+}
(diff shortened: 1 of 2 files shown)


src/components/Configure/layout/ConditionalProxyLayout.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+/**
+ * if the hydratedRevision does not have read or write,
+ * then it will not render the ConfigureInstallation
+ * @returns
+ */
+export function ConditionalProxyLayout({ children }: ConditionalProxyLayoutProps) {
+  const { hydratedRevision, loading } = useHydratedRevision();
+  const isProxy = !hydratedRevision?.content.read && hydratedRevision?.content.write;
```
laurenzlong, line 15: to make this future-proof, we should not be explicitly checking for read and write keys, because we'll have other action types in the future (subscribe, etc.). You'll probably want something like this:

```
const keys = Object.keys(hydratedRevision?.content || {})
isProxy = keys.length === 1 && keys[0] === "proxy"
```

src/components/Configure/layout/ConditionalProxyLayout.tsx
```diff
@@ -0,0 +1,25 @@
+import { LoadingIcon } from '../../../assets/LoadingIcon';
+import { useHydratedRevision } from '../state/HydratedRevisionContext';
+
+interface ConditionalProxyLayoutProps {
+  children: React.ReactNode;
+}
+
+/**
+ * if the hydratedRevision does not have read or write,
```
laurenzlong, line 9:
```suggestion
 * if the hydratedRevision only has proxy,
```

laurenzlong (approved):

src/components/Configure/layout/ConditionalProxyLayout.tsx
dionlow: will do this in a followup.

src/components/Configure/layout/ConditionalProxyLayout.tsx
dionlow: will do this in followup

src/components/Configure/layout/ConditionalProxyLayout.tsx
laurenzlong: Oh that's a great point, my idea won't work. It might make sense to keep a list in `constants.ts` for the non-proxy actions (for now, it'll be ["write", "read"]) and check the keys against that list.

PR #282, opened 2024-02-22
[ENG-733] fix: process.env.variable reference

dionlow:
### bug
process is always undefined, current fix breaks local server development. 

### solution
we add try catch block to read env or return prod url if evaluation fails. not tested in embedded client yet.

### DIFF
--- a/src/services/api.ts
+++ b/src/services/api.ts
@@ -35,15 +35,12 @@ import { ApiService } from './ApiService';
    *
    * */
 const VERSION = 'v1';
+const prodEndpoint = 'https://api.withampersand.com';
 
 function getApiEndpoint(): string {
-  const prodEndpoint = 'https://api.withampersand.com';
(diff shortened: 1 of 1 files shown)


src/services/api.ts
```diff
@@ -35,15 +35,12 @@ import { ApiService } from './ApiService';
    *
    * */
 const VERSION = 'v1';
+const prodEndpoint = 'https://api.withampersand.com';
 
 function getApiEndpoint(): string {
-  const prodEndpoint = 'https://api.withampersand.com';
+  const ENV_SERVER = process.env.REACT_APP_AMP_SERVER || 'prod';
```
RajatPawar, line 41: If `process` is undefined, won't this error out?

dionlow: sorry, yeah i forgot to remove review since realized this will still be an issue offline. updated with try catch block.

RajatPawar (approved):

PR #287, opened 2024-02-26
 [ENG-654] feat: add proxy only create installation request

dionlow:
BLOCKED: [ENG-742]

### Summary 
Part 3 of the proxy only action [ENG-654]. This PR builds on top of conditional proxy layout and now will send a createInstallation request for proxy only actions.
- update isProxyOnly control logic
- refactor create installation reducer for read, write, and proxy

#### demo
proxy-success-demo

#### Progress
1. ~~add conditional layout render for proxy support (base PR)~~
2. ~~add the `SuccessTextBox`~~
3. **add control logic for only proxy present**
4. **add the createInstallation call**

#### Followup
while proxy only actions are functional with PR... 
- proxy only error create case is not handle UI wise, only console.error is put in. 
- refactor: we see that reducer logic can be pulled out more in updateInstallation as well to not duplicate across all actions i.e. (read, write, proxy, etc). 
- create/update proxy installation is also not supported if enabled with other read/write action types.


src/components/Configure/actions/createInstallationReducer.ts
```diff
@@ -0,0 +1,49 @@
+import {
```
dionlow, line 1: this reducer is refactored into it's own file to be re-used in other actions.

src/components/Configure/layout/ConditionalProxyLayout/hasDisqualifyingKey.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * given a list of keys and disqualifying keys,
+ * return true if any of the disqualifying keys are found in the list of keys
+ * @param keys
+ * @param disqualifyingKeys
+ * @returns
+ */
+export function hasDisqualifyingKey(keys: string[], disqualifyingKeys: string[]): boolean {
+  return keys.reduce((hasDisqualifying: boolean, key: string) => {
```
laurenzlong, line 9: nit : I personally find it confusing to use `reduce` when there isn't a need to accumulate an intermediate result. This function would be more readable if implemented with `forEach` but that's just my opinion.

src/components/Configure/layout/ConditionalProxyLayout/hasDisqualifyingKey.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * given a list of keys and disqualifying keys,
+ * return true if any of the disqualifying keys are found in the list of keys
+ * @param keys
+ * @param disqualifyingKeys
+ * @returns
+ */
+export function hasDisqualifyingKey(keys: string[], disqualifyingKeys: string[]): boolean {
```
laurenzlong, line 8: This function can have a simpler name like `hasKeys` or `hasOverlap`. The fact that keys are disqualifying keys is not really relevant to the implementation this function, it just checks for whether the first set of strings contains something in the second set (in other words, whether the 2 arrays overlap/intersects). And if you rename it to be something more generic you can put it in `utils`, it might be useful for other components later.

src/components/Configure/layout/ConditionalProxyLayout/ConditionalProxyLayout.tsx
```diff
@@ -1,39 +1,68 @@
...
-  // // TMP remove read and write
-  // if (hydratedRevision) {
-  //   hydratedRevision.content.read = undefined;
-  //   hydratedRevision.content.write = undefined;
-  // }
+  const provider = hydratedRevision?.content?.provider;
 
-  const isProxy = !hydratedRevision?.content.read && !hydratedRevision?.content.write;
+  const actionKeys = Object.keys(hydratedRevision?.content || {});
+  const isProxyOnly: boolean = hasDisqualifyingKey(actionKeys, supportedNonProxyTypes);
```
laurenzlong, line 39: shouldn't this be the opposite? If the actionsKeys contains some non proxy types, then it is NOT proxy only.

src/components/Configure/layout/ConditionalProxyLayout/hasDisqualifyingKey.ts
laurenzlong: Or even simpler would be a one-liner

`return !!keys.find((key) => disqualifyingKeys.includes(key));`

laurenzlong (changes requested):

src/components/Configure/layout/ConditionalProxyLayout/ConditionalProxyLayout.tsx
dionlow: can change the naming; was named supported as ampersand support these non-proxy actions and this list would need to manually updated.

src/components/Configure/layout/ConditionalProxyLayout/hasDisqualifyingKey.ts
dionlow: updated function to be more generic.

src/components/Configure/layout/ConditionalProxyLayout/hasDisqualifyingKey.ts
dionlow: updated to be hasKeys

dionlow:
@laurenzlong - ready for re-review.

src/components/Configure/layout/ConditionalProxyLayout/ConditionalProxyLayout.tsx
```diff
@@ -1,39 +1,68 @@
...
-  // // TMP remove read and write
-  // if (hydratedRevision) {
-  //   hydratedRevision.content.read = undefined;
-  //   hydratedRevision.content.write = undefined;
-  // }
+  const provider = hydratedRevision?.content?.provider;
 
-  const isProxy = !hydratedRevision?.content.read && !hydratedRevision?.content.write;
+  const actionKeys = Object.keys(hydratedRevision?.content || {});
+  const isProxyOnly: boolean = hasKeys(actionKeys, nonProxyActionList);
```
laurenzlong, line 39: To clarify, my last comment wasn't about the naming, it's about the logic. This should be:

```
const isProxyOnly: boolean = !hasKeys(actionKeys, nonProxyActionList);
```

laurenzlong (changes requested):

src/components/Configure/layout/ConditionalProxyLayout/ConditionalProxyLayout.tsx
dionlow: could catch, thankfully this also helped find the other change with read & write keys.

dionlow:
Ready for review, but will not be confirm and merged until [ENG-742] is in.

laurenzlong (approved):

PR #298, opened 2024-03-04
[ENG-732] feat: support proxy enabled read / write

dionlow:
### Summary
We need to support proxy when read and write are configured. 8 cases of combinations to be supported
- support **read create installation** with/without proxy enabled
- support **write create installation** with/without proxy
- support **read update installation** with/without proxy
- support **write update installation** with/without proxy

#### Demo
UI should have no visible changes

#### Tested 
- locally: read/write create/update with proxy. 
- production: read/write create/update without proxy.

### DIFF
--- a/src/components/Configure/actions/read/onSaveReadCreateInstallation.ts
+++ b/src/components/Configure/actions/read/onSaveReadCreateInstallation.ts
@@ -76,6 +76,12 @@ const generateCreateReadConfigFromConfigureState = (
     },
   };
 
+  // insert proxy into config if it is enabled
+  const isProxyEnabled = hydratedRevision.content.proxy?.enabled;
+  if (isProxyEnabled) {
+    createConfigObj.content.proxy = { enabled: true };
+  }
+
   return createConfigObj;
 };
 
--- a/src/components/Configure/actions/read/onSaveReadCreateInstallation.ts
+++ b/src/components/Configure/actions/read/onSaveReadCreateInstallation.ts
@@ -9,6 +9,7 @@ import {
 } from '../../state/utils';
 import { ConfigureState } from '../../types';
 import { createInstallationAndSetState } from '../mutateAndSetState/createInstallationAndSetState';
+import { getIsProxyEnabled } from '../proxy/isProxyEnabled';
 /**
  * gets matching object from hydratedRevision
  * @param hydratedRevision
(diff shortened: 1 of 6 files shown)


src/components/Configure/actions/read/onSaveReadCreateInstallation.ts
```diff
@@ -76,6 +76,12 @@ const generateCreateReadConfigFromConfigureState = (
     },
   };
 
+  // insert proxy into config if it is enabled
+  const isProxyEnabled = hydratedRevision.content.proxy?.enabled;
```
laurenzlong, line 80: Can you put this line into a function that takes in a hydratedRevision and returns a boolean indicating whether or not proxy is enabled (we might change the format of the hydratedRevision in the future where if it has any other keys then `enabled: true` is not explicitly required). So if it was abstracted into a helper function then that logic is easy to change in the future.

src/components/Configure/actions/write/onSaveWriteCreateInstallation.ts
```diff
@@ -66,6 +66,12 @@ const generateCreateWriteConfigFromConfigureState = (
     },
   };
 
+  // insert proxy into config if it is enabled
+  const isProxyEnabled = hydratedRevision.content.proxy?.enabled;
```
laurenzlong, line 70: The added lines look fine (once `hydratedRevision.content.proxy?.enabled;` is put into a helper function), but I'm pretty confused by this `generateCreateWriteConfigFromConfigureState` function. 
- the file name is `onSaveWriteCreateInstallation` but the comment above the function refers to update installation request, is this an outdated comment?
- the comment line 59 says "need empty read.standardObjects for update read", what does this mean? Can you expand the comment line to be more clear?
- do you always add `read` and `write` keys to config.content even when the hydratedRevision doesn't have read or write actions? Have you found that things break if you don't add them?

laurenzlong (approved):

src/components/Configure/actions/write/onSaveWriteCreateInstallation.ts
dionlow: Yes, read/write was breaking on the backend if the other wasn't supplied. This may have been fixed @RajatPawar - did we have a ticket or did you fix this by chance?

dionlow: It looks like it's working now for create installation but then throws a null pointer exception during the read update. I've reported a bug for this if you did want to remove this hack. 

I've updated the comment in the code with the linear ticket for context. 
https://linear.app/ampersand/issue/ENG-780/bug-write-createupdate-installation-without-read

PR #320, opened 2024-04-04
[ENG-930, ENG-932] feat: add generic WorkspaceOauthFlow component 

dionlow:
### Summary
We want to support generic providers that require workspaces. We will keep parts of the SalesforceEntry as having a custom entry for large provider may prove useful.
- add `WorkspaceEntry` component
- adds `WorkspaceOauthFlow` and feature flag

#### demo
When we turn on the flag, salesforce will use the generic WorkspaceEntryFlow component instead of the existing custom flow. 

workspace-entry-demo

#### followup
Some of the forms can be refactored and deleted so there is less copy pasted styles.

### DIFF
--- a/src/components/Oauth/WorkspaceEntry/WorkspaceOauthFlow.tsx
+++ b/src/components/Oauth/WorkspaceEntry/WorkspaceOauthFlow.tsx
@@ -1,0 +1,80 @@
+/**
+ * OAuth flow for any providers that do not require the consumer to enter a subdomain first.
+ */
+
+import { useCallback, useState } from 'react';
+
+import { useApiKey } from '../../../context/ApiKeyContextProvider';
+import { useProject } from '../../../context/ProjectContextProvider';
+import { capitalize } from '../../../utils';
+import { fetchOAuthCallbackURL } from '../fetchOAuthCallbackURL';
+import OAuthPopup from '../OAuthPopup';
+
+import { WorkspaceEntry } from './WorkspaceEntry';
+
+interface NoSubdomainOauthFlowProps {
+  provider: string;
+  consumerRef: string;
+  consumerName?: string;
+  groupRef: string;
+  groupName?: string;
+}
+
+/**
+ * WorkspaceEntry is generic for any provider that requires a workspace to be entered first,
+ * then launches a popup with the OAuth flow.
+ */
+export function WorkspaceOauthFlow({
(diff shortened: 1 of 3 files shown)


src/components/Oauth/WorkspaceEntry/WorkspaceOauthFlow.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+import { useApiKey } from '../../../context/ApiKeyContextProvider';
+import { useProject } from '../../../context/ProjectContextProvider';
+import { capitalize } from '../../../utils';
+import { fetchOAuthCallbackURL } from '../fetchOAuthCallbackURL';
+import OAuthPopup from '../OAuthPopup';
+
+import { WorkspaceEntry } from './WorkspaceEntry';
+
+interface NoSubdomainOauthFlowProps {
```
laurenzlong, line 15: This should be `WorkspaceOauthFlowProps`

src/components/Oauth/WorkspaceEntry/WorkspaceOauthFlow.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const { projectId } = useProject();
+  const apiKey = useApiKey();
+
+  const [workspace, setWorkspace] = useState<string>('');
+  const [oAuthCallbackURL, setOAuthCallbackURL] = useState<string | null>(null);
+  const [error, setError] = useState<string | null>(null);
+
+  //  fetch OAuth callback URL from connection so that oath popup can be launched
+  const handleSubmit = async () => {
+    setError(null);
```
laurenzlong, line 39: Can you add an if statement to check that `workspace` is not an empty string, if so, call `setError('Workspace name cannot be empty)`

laurenzlong (approved):

src/components/Oauth/WorkspaceEntry/WorkspaceOauthFlow.tsx
dionlow: fixed in later pr.

src/components/Oauth/WorkspaceEntry/WorkspaceOauthFlow.tsx
dionlow: this shouldn't be possible since the submit button is disabled when the workspace is empty, but I can add in this check in the final polish pr.

PR #330, opened 2024-04-05
[ENG-726] ci: check that the project is buildable

lino-levan:
In #272, code that did not build (outside of my machine) managed to land in main because we didn't check that the code was buildable in the CI environment. This PR simply checks that the code is able to build correctly during linting.


dionlow (approved):
thanks for adding!

PR #338, opened 2024-04-11
[ENG-731] fix: styling issue with text color on white background

lino-levan:
When embedders use @amp-labs/react, the styling of their website leaked into the components in such a way that text could appear invisible (white-on-white). 

Before:
image
image

After:
image
image


lino-levan:
The answer to my question is: yes. Mailmonkey repro:

image

Let me fix those cases as well.

dionlow (approved):
LGTM. nice catch!

PR #346, opened 2024-04-18
fix: allow groupName and consumerName to be optional

dionlow:
### Summary
When component did not have `consumerName` the app failed to launch the Oauthflow. We remove the if checks and go to the try catch so errors will be handled in the catch loop. We also make `groupName` optional. 

#### testing
Screenshot 2024-04-18 at 4 49 06 PM


laurenzlong (approved):
I assume you've tested the whole flow to ensure no other parts broke when consumerName or groupName is missing?

dionlow:
Yeah, I walked through the flow earlier. I can also do a quick walk through before release.

PR #367, opened 2024-05-10
[bugfix] add isLoading to condintion

jlimatampersand:
Add !isLoading condition in if check to call createInstallation to prevent multiple calls being made. 

Screenshot 2024-05-10 at 4 23 28 PM


laurenzlong (approved):
Nice! Can you actually add `src/services/version.ts` to .gitignore? (But make sure the file doesn't get deleted) ,we want the file to be in the repo, but the changes to it should't be checked in).

PR #369, opened 2024-05-11
[chore] Remove git ignore version.ts

jlimatampersand:
As per discussion in here: https://github.com/amp-labs/react/pull/367#pullrequestreview-2050977646

### DIFF
--- a/src/services/.gitignore
+++ b/src/services/.gitignore
@@ -1,0 +1,1 @@
+version.ts


laurenzlong:
I actually already did it in https://github.com/amp-labs/react/pull/368/files (sorry, I didn't know if you saw my comment or not since you didn't ack it)

PR #376, opened 2024-05-22
[ENG-1136] feat: rework oauth flow, add better error handling

dionlow:
## Summary
This PR reworks the Oauth flow to be more modular and easier to maintain. 
- adds OAuthWindow component
- adds OAuthWindowToast components (for success and errors)
- refactors window/message handling into functions
- swaps out the OAuthPopup for OAuthWindow in NoWorkspaceEntry only. (does not do salesforce/workspace entry yet)

The PR rewrites the Oauth flow to help to debug the cross-origin frame issue. 
### Followup
- delete depreciated OAuthPopup 
- update Salesforce/WorkspaceEntry flow.

#### Potential Followup
It looks like this solution is not perfect. There is still some window manipulation or mishandling of event listeners happening. The PR works with the current backend auto-close, but finding a standardize way to handle the windows seems more difficult than it should be. 
1. We can merge this PR as is and re-visit if there are more issues
2. Need another set of eyes to re-think what's happening with window manipulation (global this object, window instance, window references etc)

## Problem 
Occasionally the user will attempt to authorize and there will be a cross-origin frame error which may occur in either a Success or Failure event. To make the problem more difficult to debug includes
- the use of windows and manipulation of windows via both backend and frontend
- the use of timeouts, intervals, and messages being constantly polled
- window, timeouts, and references make the code difficult to follow and debug

### Initial Approach
1. Attempt to debug by forcing and reproducing the error
3. Reproducing was not always consistent event with hard-coded server errors -> attempt to re-write the code in an easier to understand, maintain, and modular way.
4. Replace NoWorkplaceOauthFlow with new OAuthWindow component and deprecate OauthPopup.
5. Bug still occurs, attempt to debug.

### Root Cause Hypothesis.
Since forcing an error and refactoring the code did not remove the error, the error might be related to how the backend sends information to the front-end. Upon inspection, it seems that one of root causes could be related to the manipulation of windows.

1. The popup will create a window and open the window. 
2. The front-end will also create a message listener.
3. The backend sends a failure or success message and closes the window. 
4. The front-end clears the interval timer and relays a onClose message to the parent form.
6. User either goes to next page or fills out the form and returns to step 1.

### The problem continues
**What happens if the authorization fails or the user exits the popup?**
The user must attempt to login again. 

**When the user logs in, a new window is created; but what happened to the previous window and the previous window listener?** 
The backend has sent a message to close the window, but the window still exists and step 1 creates a new window even though the previous window can still be opened. Now if authorization fails, the user closes the window, or the backend closes the window...we'll eventually have a bunch of windows and window listeners listening for the same event.

 If one of the previous windows hears the message from the backend success or fail, it's possible this may qualify as a cross-origin frame error as the other windows are being manipulated when not intended.

## The potential solution
1. Create a reference to open and reopen the same window and attach and re-attach the message listener
2. Skip the reference, fully clean up the window and the event listener.

This PR does the 2nd option to avoid reference complexity. Whenever the window is closed, both the current window and window listeners are removed or set to null.

#### before
cross-origin-frame
 
#### after
oauth-flow-update


src/components/Oauth/NoWorkspaceEntry/NoWorkspaceOauthFlow.tsx
```diff
@@ -49,7 +49,7 @@ export function NoWorkspaceOauthFlow({
       );
       setOAuthCallbackURL(url);
     } catch (err: any) {
-      console.error(err);
+      console.error('Could not fetch OAuthCallback URL', { err });
```
laurenzlong, line 52: I realized that this ` fetchOAuthCallbackURL` was actually a poorly named function, since it can be confused with the callback URL that that Salesforce redirects to. Can you rename it `fetchOAuthPopupURL` or something like that, and revise this error message to say 'Could not fetch OAuth popup URL'

src/components/Oauth/OAuthWindow/OAuthWindow.tsx
```diff
@@ -0,0 +1,79 @@
...
+  onClose: (err: string | null) => void;
+};
+
+/**
+ * v2 OAuth window to manage OAuth flow
+ * @param param0
+ * @returns
+ */
+export function OAuthWindow({
+  children, oauthUrl, windowTitle = 'OAuthWindow', onClose,
```
laurenzlong, line 23:
```suggestion
  children, oauthUrl, windowTitle = 'Connect to Provider', onClose,
```

src/components/Oauth/OAuthWindow/windowHelpers.tsx
```diff
@@ -0,0 +1,95 @@
...
+const FAILURE_EVENT = 'AUTHORIZATION_FAILED';
+
+/**
+ * returns a function to refresh the connection
+ * @param projectId
+ * @param apiKey
+ * @param setSelectedConnection
+ * @returns
+ */
+export function refreshConnection(
```
laurenzlong, line 20: All of the functions in this helper file are higher order functions that return functions, it would be more obvious if they were named like:
- getRefreshConnectionHandler
- getOpenWindowHandler
- getMessageHandler

src/components/Oauth/OAuthWindow/OAuthWindow.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }, [oauthUrl, oauthWindow, openOAuthWindow, receiveMessage, windowTitle]);
+
+  // refresh connections on connectionId change
+  useEffect(() => {
+    if (connectionId) {
+      refreshConnections(connectionId)
+        .then(() => {
+          oauthWindow?.close(); // only close the window if connection is successful
+          // show success toast if connection is successful
+          ToastSuccessConnection();
```
laurenzlong, line 50: We shouldn't do a toast, that is too intrusive. Toasts are ok to do in our own console, but not in the builder's app since we have no idea what their layout is, and we should keep our UI within the embeddable component itself & the oauth popup. So please remove `ToastSuccessConnection`, `ToastOauthFailed` and all the other toast related code.

src/components/Oauth/OAuthWindow/OAuthWindow.tsx
```diff
@@ -0,0 +1,79 @@
...
+      // cleanup event listener and window reference
+      window.removeEventListener('message', receiveMessage);
+      setOauthWindow(null);
+
+      if (!connectionId) {
+        // if connectionId is not set, then set OAuth failed error
+        ToastOauthFailed();
+        if (onClose) onClose('OAuth failed. Please try again.');
+      } else if (connectionId && onClose) {
+      // if connectionId is set, then set OAuth success -- no error in
```
laurenzlong, line 72: this seems to be an incomplete sentence

laurenzlong (approved):
Thanks for the thorough explanation!

PR #419, opened 2024-07-11
build: update build process to use vite library mode

dionlow:
### Summary
This PR adds vite as a build tool.
- adds vite dependency
- vite define config file
- updates package.json to use vite commands to compile into CommonJs and ES module files.
- adds ts types into a single export file.

Our library build support relies mostly on basic eslint, babel, and typescript to compile down into common js to be most compatible. One of the downsides is not being able to leverage some of the debugging capabilities and ease of readability in ES module syntax. This PR primarily grants support for both. Vite will also provide additional tooling support around packages and aliases which are not supported yet in these repo.

#### Resources
vite library mode: https://vitejs.dev/guide/build#library-mode
vite supports rollup bundling: https://rollupjs.org/ 

#### Unknowns
- updates to the code seems to break on refresh and mail-monkey needs to be restarted. (mailmonkey fix in progress)
- there may be breaking changes since compile process is different; should be reflected in version number.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,15 +15,20 @@
     "registry": "https://registry.npmjs.org/"
   },
   "homepage": "https://github.com/amp-labs/react#readme",
-  "main": "build/src/index.js",
-  "typings": "build/src/index.d.ts",
-  "files": [
-    "/build"
-  ],
+  "typings": "./build/index.d.ts",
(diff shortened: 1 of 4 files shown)


laurenzlong (approved):
Tried it out locally and `yarn build` and `yarn watch` both work for me.

PR #425, opened 2024-07-15
build: add path alias for vite to tsconfig

dionlow:
### Summary
- adds alias for generated-sources
- adds alias for components, hooks, services, context

Adds alias for non-component folders. 

### Followup 
Will add aliases to component folder.


RajatPawar (approved):
Cool!

PR #462, opened 2024-08-08
[ENG-1407] fix: allow required only reads to be saved

dionlow:
### Summary
The current configure component requires a dirty state or create state to enable submit. This PR adds a check to see if the object has been completed, if not completed as a read state, it will be considered a new state so disable will not be activated
- adds check if selected object is completed, before enabling disable.
#### testing

`amp.yaml`
```
specVersion: 1.0.0
integrations:
  - name: readHubspotContactsRequiredOnly2
    displayName: Read Hubspot Contacts
    provider: hubspot
    read:
      objects:
        - objectName: contact
          destination: defaultWebhook
          schedule: "*/30 * * * *" # Every 30 minutes
          requiredFields:
            # First required field is "name"
            - fieldName: name    
        - objectName: account
          destination: defaultWebhook
          schedule: "*/30 * * * *" # Every 30 minutes
          requiredFields:
            # First required field is "name"
            - fieldName: name
  ```

##### before
Screenshot 2024-08-08 at 6 24 45 PM

##### after
enable-require-only-saves

#### todo 
tests working with other optional read/ write configurations

### DIFF
--- a/src/components/Configure/content/ConfigureInstallationBase.tsx
+++ b/src/components/Configure/content/ConfigureInstallationBase.tsx
@@ -6,10 +6,12 @@ import {
 } from '@chakra-ui/react';
 
 import { LoadingIcon } from 'assets/LoadingIcon';
+import { useInstallIntegrationProps } from 'context/InstallIntegrationContextProvider';
 
 import { OTHER_CONST } from '../nav/ObjectManagementNav/constant';
 import { UNINSTALL_INSTALLATION_CONST } from '../nav/ObjectManagementNav/UninstallInstallation';
 import { useHydratedRevision } from '../state/HydratedRevisionContext';
+import { getReadObject } from '../utils';
 
 import { ReadFields } from './fields/ReadFields';
 import { WriteFields } from './fields/WriteFields';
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by dionlow** [comment]
@laurenzlong - tagged you since the UX is an alternative solution to what is proposed in the ticket.

File: src/components/Configure/content/ConfigureInstallationBase.tsx [hunk 1]
```diff
@@ -29,17 +31,24 @@ export function ConfigureInstallationBase(
     onSave, onReset, isLoading, isCreateMode = false,
   }: ConfigureInstallationBaseProps,
 ) {
+  const { installation } = useInstallIntegrationProps();
   const { hydratedRevision, loading } = useHydratedRevision();
   const { configureState, selectedObjectName } = useSelectedConfigureState();
 
+  // check if selected object is completed.
```
laurenzlong, line 38: Can you add that, in the case the object does not have any configuration required, it is also considered "complete"?

laurenzlong (approved):
This is a fine solution for now, if it's not too difficult to have the button say "Enable" instead of "Save" in the case that the object does not have any configuration required, that would be less confusing for consumers (it's weird to see "Save" when I haven't done anything)

src/components/Configure/content/ConfigureInstallationBase.tsx
dionlow: The code doesn't check for required fields. The code checks if the object exists in the config. If the object doesn't exists, we assume it is not completed by the user yet, therefore validation is not enforced.

laurenzlong:
@dionlow I'm going to merge this now so I can release it, but if my suggestion above isn't too difficult can you make a follow up PR?

PR #466, opened 2024-08-14
feat: Add a builder image

eberle1080:
This introduces a builder image. This is necessary to create automations, such as a cloudbuild trigger to automatically update generated sources with the latest openapi definitions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 FROM node:22-alpine3.19
 SHELL ["/bin/ash", "-o", "pipefail", "-c"]
 WORKDIR /app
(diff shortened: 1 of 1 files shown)


dionlow:
It's possible that automatically updating generated sources could break react when not expected (since endpoints can be deprecated, or fields changed). We might still want to keep the manual verification as a separate PR, but open to discussion.

eberle1080:
@dionlow "automatic" may be a bit of an oversimplification on my part. It's automated in the sense that it reacts to changes in the openapi repo by opening a PR. There will still be an approval / merge step by an actual human. It just automates most of the busy work.

dionlow (approved):

PR #488, opened 2024-08-23
[ENG-1350] fix: bring back more granular error handling for auth errors

dionlow:
### Summary 
Bring back granular error handling. 
- renames onClose to onError since we can perform errors without closing the window.

note: did not replicate, will need help to set up local env to mimic this error.

### DIFF
--- a/src/components/Oauth/NoWorkspaceEntry/NoWorkspaceOauthFlow.tsx
+++ b/src/components/Oauth/NoWorkspaceEntry/NoWorkspaceOauthFlow.tsx
@@ -55,7 +55,7 @@ export function NoWorkspaceOauthFlow({
     }
   };
 
-  const onClose = useCallback((err: string | null) => {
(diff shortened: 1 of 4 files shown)


laurenzlong:
I checked this branch out locally and verified that the successful oauth flow still works, but when I tried to trigger an error, the raw HTML string gets shown like this:
Screenshot 2024-08-28 at 2 24 31 PM

I'm pretty confused by why this is happening, I thought it was maybe a server issue, so I created https://github.com/amp-labs/server/pull/1591 but that still didn't work. I won't have time to work more on this before going OOO, so let's pause on this for now and come back to it later.

laurenzlong (approved):
Let's actually just merge this - since it makes the logic cleaner and the success flow still works, and it does not make the failure flow worse. The issue in the screenshot happens on main branch also.

PR #542, opened 2024-09-23
chore: remove unused API services

laurenzlong:
We'll be introducing frontend-only API keys soon, which will have less access than full API keys. This PR removes unnecessary API services from from being exported.


dionlow (approved):
lgtm!

PR #552, opened 2024-09-26
[ENG-1561] feat/refactor: salesforce subdomain entry component

dionlow:
### Summary
refactor SalesforceSubdomainEntry to not use Chakra. Adds bridge Component to swap out Chakra implementation.
- add native button
- add native input
- mv salesforce subdomain entry to chakra file (deprecated)

#### demo

salesforce-domain-refactor

#### previous
Screenshot 2024-09-26 at 2 31 02 PM

#### after
Screenshot 2024-09-26 at 2 31 26 PM

### DIFF
--- a/src/components/auth/Oauth/Salesforce/SalesforceSubdomainEntry.tsx
+++ b/src/components/auth/Oauth/Salesforce/SalesforceSubdomainEntry.tsx
@@ -1,59 +1,47 @@
-import { ExternalLinkIcon } from '@chakra-ui/icons';
-import {
-  Button, FormControl,
-  FormLabel, Heading, Input,
-} from '@chakra-ui/react';
-
 import { AuthErrorAlert } from 'components/auth/AuthErrorAlert/AuthErrorAlert';
+import { FormComponent } from 'components/form';
 import { AccessibleLink } from 'components/ui-base/AccessibleLink';
+import { Button } from 'components/ui-base/Button';
+import { isChakraRemoved } from 'src/components/ui-base/constant';
 import { AuthCardLayout } from 'src/layout/AuthCardLayout/AuthCardLayout';
 
-const SALESFORCE_HELP_URL = 'https://help.salesforce.com/s/articleView?id=sf.faq_domain_name_what.htm&type=5';
+import { ChakraSalesforceSubdomainEntry } from './ChakraSalesforceSubdomainEntry';
+import { SALESFORCE_HELP_URL, SubdomainEntryProps } from './SubdomainEntryProps';
 
-type SubdomainEntryProps = {
-  handleSubmit: () => void;
-  setWorkspace: (workspace: string) => void;
-  error: string | null;
-  isButtonDisabled?: boolean;
-};
-
-/**
- * Salesforce specific subdomain entry component, use workspace entry for other providers.
- * @param param0
- * @returns
- */
 export function SalesforceSubdomainEntry({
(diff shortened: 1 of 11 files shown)


src/components/ui-base/Button/button.module.css
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.button:disabled {
+    background-color: var(--amp-colors-button-primary-disabled);
+    cursor: not-allowed;
+}
+
+/* Additional styling to support screen readers */
+button[aria-expanded="true"] {
+    background-color: #004080;  /* Can visually indicate button state */
```
RajatPawar, line 33: Would be good to export this as a var too

src/components/form/Input/input.module.css
```diff
@@ -0,0 +1,36 @@
...
+    outline: none;
+}
+
+.input:disabled {
+    background-color: var(--amp-colors-neutral-300);
+    border: 1px solid var(--amp-colors-neutral-400);
+}
+
+.error {
+    color: var(--color-red-600, #DC2626);
```
RajatPawar, line 32: Same as the other comment

src/components/auth/Oauth/Salesforce/SalesforceSubdomainEntry.tsx
```diff
@@ -1,59 +1,47 @@
...
+      <h1 style={{ fontWeight: 600, lineHeight: 1.2, fontSize: '1.2em' }}>Enter your Salesforce subdomain</h1>
+      <AccessibleLink href={SALESFORCE_HELP_URL} newTab>
+        What is my Salesforce subdomain?
+      </AccessibleLink>
+      <AuthErrorAlert error={error} />
+      <div style={{ display: 'flex', marginTop: '1em' }}>
+        <FormComponent.Input
+          id="salesforce-subdomain"
+          type="text"
+          placeholder="MyDomain"
```
RajatPawar, line 32: I would lowercase this because subdomains would generally look like `my-domain` or `mydomain`

src/components/form/Input/index.tsx
```diff
@@ -0,0 +1,27 @@
+import { InputHTMLAttributes } from 'react';
+import classNames from 'classnames';
+
+import classes from './input.module.css';
+
+export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
```
RajatPawar, line 6: What do you think about allowing a validator here? For example, salesforce subdomains are supposed to be all lowercase, between 4-19 characters, etc ([ref](https://help.salesforce.com/s/articleView?id=sf.mth_set_up_your_mth_subdomain_and_namespace.htm&type=5)), and we could save a roundtrip for the user if we allow pre-API validation. The optionality to have a validator do that would be nice

RajatPawar (approved):

src/components/form/Input/index.tsx
dionlow: Yeah inline-validation and advanced form state (isDirty, isTouched etc) is out of scope right now given the size of the refactor/rewrite. Definite consideration in the future tho.

src/components/form/Input/input.module.css
dionlow: hmm yeah, this style isn't used yet since we don't support advanced form state yet. Will clean up in a followup pr.

src/components/ui-base/Button/button.module.css
dionlow: same as above.

PR #570, opened 2024-09-30
[ENG-1561] refactor: basic auth flow

dionlow:
### Summary
adds native implementation of basic auth flow and bridge component

#### screenshot
Screenshot 2024-09-30 at 4 49 19 PM


src/components/auth/BasicAuth/BasicAuthFlow.tsx
```diff
@@ -60,14 +49,15 @@ export function BasicAuthFlow({
   }, [apiKey, provider, nextStep, consumerName, consumerRef, groupName,
     groupRef, project, creds, setSelectedConnection]);
 
-  const onNext = (user: string, pass: string) => {
-    setCreds({ user, pass });
+  const onNext = (form: any) => {
```
jatinsandilya, line 52: nit: the types could be tigher here IMO instead of using `any`

jatinsandilya (approved):
Added a comment on types otherwise looks good

PR #606, opened 2024-10-15
[ENG-1600] feat: support reinstall integration

dionlow:
### Summary 
Users are asked to delete and re-install integration when connection id or something is not set correctly. We need to support reinstallation in the UI lib.
- adds button when installation is deleted. 
- reinstall button will trigger a seed refresh and force the InstallIntegration component to re-render. 

As a side note, this logic can be forced during the onUninstallIntegration callback in the parent app. 

note: this is hidden behind feature flag so we don't have to support using a ChakraButton.

### Demo
re-install-integration

### DIFF
--- a/src/components/Configure/InstallIntegration.tsx
+++ b/src/components/Configure/InstallIntegration.tsx
@@ -1,3 +1,5 @@
+import { useState } from 'react';
+
 import { ErrorTextBox } from 'components/ErrorTextBox/ErrorTextBox';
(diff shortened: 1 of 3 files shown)


eberle1080 (approved):
Interesting approach. So by re-seeding this value, it forces a re-render?

dionlow:
Yeah sort of like a game seed. When the seed changes, the original component will render the new seed.

PR #610, opened 2024-10-16
[ENG-1610] feat: add env ff to remove chakra-ui

dionlow:
### Summary
To test this in Mailmonkey inject 
```
REACT_APP_IS_CHAKRA_REMOVED=true
```

into env variable or start script like

```json
"start:prod": "PORT=3001 REACT_APP_AMP_SERVER=prod REACT_APP_IS_CHAKRA_REMOVED=true npm start",
```

### DIFF
--- a/src/components/ui-base/constant.ts
+++ b/src/components/ui-base/constant.ts
@@ -12,4 +12,4 @@
  */
 
 // parent app my turn on or off the chakra-ui components with this flag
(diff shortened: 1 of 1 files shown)


laurenzlong (approved):
Can you rename this flag something like `REACT_APP_ENABLE_CSS_MODULE` (so it's a more positive sounding env variable name, and conveys the feature that it unlocks)

dionlow:
renamed REACT_APP_ENABLE_CSS since the builder won't be able to necessarily override with module css, but can override with basic css.

PR #612, opened 2024-10-17
[ENG-1602] feat: add `fieldMapping` prop to `InstallIntegration` component

jatinsandilya:
Adds dynamic fieldMapping prop to `InstallIntegration` component so builders can add dynamic field mappings per consumer. 

Tested it manually with `demo-hubspot-apollo-app`: https://github.com/amp-labs/demo-hubspot-apollo/pull/1 

How the mapping UI looks like when both static & dynamic fieldMappings are present: 

Screenshot 2024-10-18 at 6 28 36 PM

How the mapping UI looks like when both static & dynamic fieldMappings have to map the same field: 

Screenshot 2024-10-18 at 7 17 32 PM

Tested saving the configuration as well: 

Screenshot 2024-10-18 at 7 22 51 PM

### DIFF
--- a/src/components/Configure/InstallIntegration.tsx
+++ b/src/components/Configure/InstallIntegration.tsx
@@ -5,7 +5,7 @@ import { ConnectionsProvider } from 'context/ConnectionsContextProvider';
 import { ErrorBoundary, useErrorState } from 'context/ErrorContextProvider';
 import { InstallIntegrationProvider } from 'context/InstallIntegrationContextProvider';
 import { useProject } from 'context/ProjectContextProvider';
-import { Config } from 'services/api';
+import { Config, IntegrationFieldMapping } from 'services/api';
 
 import { InstallationContent } from './content/InstallationContent';
 import { ConditionalProxyLayout } from './layout/ConditionalProxyLayout/ConditionalProxyLayout';
@@ -23,25 +23,7 @@ function useForceUpdate() {
   return { seed, reset };
 }
 
-interface Mapping {
-    /**
-     * The name of the field to map to.
-     * Example: "source"
-     */
-    mapToName: string;
-    /**
-     * The display name of the field to map to.
-     * Example: "Source"
-     */
-    mapToDisplayName?: string;
-    /**
-     * A prompt to display to the user for this field.
-     * Example: "Which field do you use to track the source of a contact?"
-     */
-    prompt?: string;
-}
-
-export type FieldMappingsType = Map<string, Array<Mapping>>
+export type FieldMappingsType = Map<string, Array<IntegrationFieldMapping>>
 
 interface InstallIntegrationProps {
   /**
--- a/src/components/Configure/InstallIntegration.tsx
+++ b/src/components/Configure/InstallIntegration.tsx
@@ -23,7 +23,7 @@ function useForceUpdate() {
   return { seed, reset };
 }
 
-export type FieldMappingsType = Map<string, Array<IntegrationFieldMapping>>;
+export type FieldMappingsType = { [key: string]: Array<IntegrationFieldMapping> };
 
 interface InstallIntegrationProps {
   /**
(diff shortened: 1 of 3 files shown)


src/components/Configure/content/fields/FieldMappings/RequiredFieldMappings.tsx
```diff
@@ -31,10 +34,23 @@ export function RequiredFieldMappings() {
   };
 
   const integrationFieldMappings = useMemo(
-    () => configureState?.read?.requiredMapFields?.filter(
-      isIntegrationFieldMapping,
-    ) || [],
-    [configureState],
+    () => {
+      if (!selectedObjectName || !fieldMapping) return [];
+      const dynamicIntegrationFieldMappings = fieldMapping ? Object.values(fieldMapping[selectedObjectName] || {}).flat() : [];
```
dionlow, line 39: nit: eslint code style says a bit too long; break into a few lines for readability.

src/components/Configure/content/fields/FieldMappings/RequiredFieldMappings.tsx
```diff
@@ -31,10 +34,23 @@ export function RequiredFieldMappings() {
   };
 
   const integrationFieldMappings = useMemo(
-    () => configureState?.read?.requiredMapFields?.filter(
-      isIntegrationFieldMapping,
-    ) || [],
-    [configureState],
+    () => {
+      if (!selectedObjectName || !fieldMapping) return [];
```
dionlow, line 38: Can you add some comments here on this transformation to make this easier to maintain?

dionlow:
Can you also test this with Chakra removed? It looks like options are not being pre-filled in the native Combobox.  

Screenshot 2024-10-21 at 2 33 29 PM

dionlow:
This looks pretty solid. Left a few comments and will need to test when Chakra is removed during this transition phase.

dionlow:
To test with Chakra removed. 
1) Set env variable in parent app (builder app/ mailmonkey) 
```REACT_APP_ENABLE_CSS=true```

You can also inject it directly in the start script for example:
```
 "start:prod": "PORT=3001 REACT_APP_ENABLE_CSS=true REACT_APP_AMP_SERVER=prod npm start",
```

2) Add CSS import at top of builder app. 
```js
import { AmpersandProvider } from '@amp-labs/react';
import '@amp-labs/react/styles';
import './App.css'; // your local css override (in development)

const options = {
  project: 'PROJECT', // Your Ampersand project name or ID.
  apiKey: 'API_KEY',// Your Ampersand API key.
};

function App() {
  return (
    // Wrap all your components inside AmpersandProvider.
    // You can either do this at the App level,
    // or further down in the component tree.
    <AmpersandProvider options={options}>
        // You can use any of the Ampersand components here.
        ...
          <InstallIntegration 
            integration = {myIntegrationName}
            consumerRef = {userId}
            consumerName = {userFullName}
            groupRef = {teamId}
            groupName = {teamName}
          />
    </AmpersandProvider>
  )
}
```

RajatPawar (approved):

dionlow:
My bad: I forgot to mention that the feature flag was broken and was merged in just before this PR: https://github.com/amp-labs/react/pull/620

I can create another [ticket](https://linear.app/ampersand/issue/ENG-1620/dynamic-field-mappings-no-chakra) to track this issue with native combobox not working with dynamic field mappings.

src/components/Configure/InstallIntegration.tsx
```diff
@@ -23,12 +23,34 @@ function useForceUpdate() {
   return { seed, reset };
 }
 
+export type FieldMappingsType = { [key: string]: Array<IntegrationFieldMapping> };
```
laurenzlong, line 26: @jatinsandilya Can you rename this to `FieldMappings`? We shouldn't post-pend "Type" at the end of types, it's unnecessarily verbose. Also you'll have to explicitly re-export it in https://github.com/amp-labs/react/blob/main/src/public/index.ts (and that file is exported in https://github.com/amp-labs/react/blob/main/src/index.ts). 

So then builders can directly use the type in their code like so:

```
import { FieldMappings, InstallIntegration } from '@amp-labs/react';
```

PR #613, opened 2024-10-17
fix/build: surround ff env var in try/catch

dionlow:
### Summary
fix build by surrounding process.env in try catch

### DIFF
--- a/src/components/ui-base/constant.ts
+++ b/src/components/ui-base/constant.ts
@@ -1,3 +1,5 @@
+import { getEnvVariable } from 'src/utils';
+
 /**
  * feature flag to remove chakra-ui
  * components should work with and without chakra-ui, and the imported styles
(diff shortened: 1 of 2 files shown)


src/components/ui-base/constant.ts
```diff
@@ -12,4 +12,13 @@
  */
 
 // parent app my turn on or off the chakra-ui components with this flag
-export const isChakraRemoved = process.env.REACT_APP_ENABLE_CSS || false;
+
+function getEnvVariable(varName: string) {
```
laurenzlong, line 16: a more optimized approach would be to add a second param called "default", and then move this function into a utility file so it can be reused in the future, optionally refactor api.ts to use it.

eberle1080 (approved):

laurenzlong (approved):

src/components/ui-base/constant.ts
dionlow: good suggestions, will refactor into a util with default.

PR #631, opened 2024-10-23
[ENG-1620] feat: rename field mapping type & add public export

jatinsandilya:
1. Renames `fieldMappingType` to `fieldMapping`
2. Adds a public export of this type
3. Ensures combobox is pre-filled without chakra
4. Ensures options are loaded on load and not just while typing

Screenshot 2024-10-23 at 5 15 49 PM

Demo video: https://www.loom.com/share/36793eac1ea642659ea47ab548eeaeb0

### DIFF
--- a/src/components/Configure/InstallIntegration.tsx
+++ b/src/components/Configure/InstallIntegration.tsx
@@ -23,7 +23,7 @@ function useForceUpdate() {
   return { seed, reset };
 }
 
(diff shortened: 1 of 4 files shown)


dionlow:
combobox still doesn't work the page is loaded an no value set; might be an issue with the Combobox. I'll merge this and create a followup.

dionlow (approved):
Will do a followup to look at combobox.

PR #643, opened 2024-10-25
[ENG-1599] fix: oauth error flashing before a successful connection

jatinsandilya:
This PR fixes the OAuth failed error that flashes before a successful connection. 

**Root cause:** Even when the connection is successful, the setInterval call was being multiple times due to re-renders. I wrapped it under a `useEffect` so it is now only called when the window `closed` status changes and no new intervals are created on re-renders. 

**Debugging:** 

Screenshot 2024-10-25 at 12 38 21 PM

Screenshot 2024-10-25 at 1 12 44 PM
Screenshot 2024-10-25 at 1 13 40 PM

**Test Screenshots (post-fix):**

Screenshot 2024-10-25 at 1 36 57 PM

Screenshot 2024-10-25 at 1 38 48 PM

**(on actual error)**


dionlow (approved):
Good find mate!

PR #670, opened 2024-11-06
fix: field mappings for required fields

dionlow:
### Summary
Required field mappings was returning an empty array if fieldMappings didn't exists. We update the code to remove this line (fix), and to separate extracting required map fields and dynamic field mappings to be easier to read and maintain. 

#### only required fields
Screenshot 2024-11-06 at 11 41 46 AM

#### with required fields and dynamic field
Screenshot 2024-11-06 at 11 42 23 AM

### DIFF
--- a/src/components/Configure/content/fields/FieldMappings/RequiredFieldMappings.tsx
+++ b/src/components/Configure/content/fields/FieldMappings/RequiredFieldMappings.tsx
@@ -36,24 +36,24 @@ export function RequiredFieldMappings() {
   };
 
   const integrationFieldMappings = useMemo(() => {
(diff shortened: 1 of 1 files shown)


src/components/Configure/content/fields/FieldMappings/RequiredFieldMappings.tsx
```diff
@@ -36,25 +36,28 @@ export function RequiredFieldMappings() {
   };
 
   const integrationFieldMappings = useMemo(() => {
-    if (!selectedObjectName || !fieldMapping) return [];
```
dionlow, line 39: this line was the root cause.

laurenzlong (approved):

PR #712, opened 2024-11-20
ux: improve ComboBox ux

dionlow:
### Summary
Improve the Combobox UX. Previously the combo only persisted typed state and always filtered the selected state so the full option list didn't show when returning to pick a different option. 

**Changes**
- show list when user clicks on input
- filters only on type
- deselects component and persists a selected state instead of typed input

combobox-ux-improvement


laurenzlong (approved):
Nice!

PR #735, opened 2024-12-06
fix: move ampersand provider errors to component level

dionlow:
### Summary
This continues work on Ampersand Provider and removes error handling from the Provider Level and moves to component levels 
- moves error handling to InstallIntegration and ConnectProvider

### DIFF
--- a/src/context/ProjectContextProvider.tsx
+++ b/src/context/ProjectContextProvider.tsx
@@ -2,7 +2,6 @@ import {
   createContext, useContext, useEffect, useMemo, useState,
 } from 'react';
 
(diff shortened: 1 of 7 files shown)


src/context/ProjectContextProvider.tsx
```diff
@@ -74,13 +73,5 @@ export function ProjectProvider(
     isLoading,
   }), [projectIdOrName, project, isLoading]);
 
-  return (
```
dionlow, line 77: removes error from ProjectProvider (inside AmpersandProvider)

PR #762, opened 2024-12-18
[ENG-1670] refactor: move api key to API service config

dionlow:
### Summary
- create a useAPI hook like in the client
- refactor `project` api call to use new service

#### not included / followup 
- migrating all instances of api 
- setting / overriding api key service instance

#### data fetched as expected
Screenshot 2024-12-18 at 12 19 22 PM

### DIFF
--- a/src/services/api.ts
+++ b/src/services/api.ts
@@ -1,5 +1,6 @@
 /* eslint-disable import/no-relative-packages */
 // currently not using a bundler to support alias imports
+import { useCallback } from 'react';
 import {
   BackfillConfig,
   Config,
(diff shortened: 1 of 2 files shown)


jatinsandilya (approved):

src/services/api.ts
```diff
@@ -100,8 +103,39 @@ export const setApi = (api: ApiService) => {
...
 export const api = () => apiValue;
 
+/**
+ * hook to access the API service
+ * @returns
+ */
+export function useAPI(): () => Promise<ApiService> {
+  const apiKey = useApiKey();
+
+  const getAPI = useCallback(async () => {
```
laurenzlong, line 118: Maybe I'm missing something but I don't understand why this needs to be async, nothing inside of this function is async, and `useAPI` becomes much harder to use if it returns a promise instead of just returning an ApiService

laurenzlong (changes requested):

laurenzlong (approved):
Thanks! Can you make a comment in useAPI that even though it doesn't need to be be async right now, we want to be able to support other ways to authenticating to the API in the future which may require async operations?

PR #765, opened 2024-12-19
[ENG-1758]feat: support OptionalFieldMappings

dionlow:
### Summary 
Supports optional field mappings
- creates a separate section for optional field mappings

#### does not test
- with `lead` object -- hydrated revision error
- with dynamic field mappings

further testing maybe necessary 

#### demo 
- test optional only
test-optional-map-fields


laurenzlong:
Tested for dynamic field mapping + having optional field mappings co-exist with optionalMappingAuto
Screenshot 2024-12-19 at 2 00 41 PM
Screenshot 2024-12-19 at 11 56 58 AM

laurenzlong (approved):

PR #781, opened 2025-01-03
[ENG-1777] fix: bug where selectedConnection wasn't reset even when provider changed

laurenzlong:
See [linear ticket ](https://linear.app/ampersand/issue/ENG-1777/integration-gets-installed-with-the-wrong-providers-connection) for details.

I verified that regular install/update still works via MailMonkey:

Screenshot 2025-01-03 at 3 14 15 PM


src/context/ConnectionsContextProvider.tsx
```diff
@@ -84,18 +84,25 @@ export function ConnectionsProvider({
         .then((_connections) => {
           setLoadingState(false);
```
laurenzlong, line 85: this is the only file that changed, everything else is autogenerated

jlimatampersand (approved):

PR #786, opened 2025-01-07
[ENG-1780] fix: reinstall proxy flow (part 2)

dionlow:
### Summary
- add a new step to manually install proxy integration
- removes automatic proxy installation up connection detection

#### demo

proxy-reinstallation-flow

### DIFF
--- a/src/components/Configure/layout/ConditionalProxyLayout/ConditionalProxyLayout.tsx
+++ b/src/components/Configure/layout/ConditionalProxyLayout/ConditionalProxyLayout.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useCallback, useState } from 'react';
 
 import { useApiKey } from 'context/ApiKeyContextProvider';
 import { useConnections } from 'context/ConnectionsContextProvider';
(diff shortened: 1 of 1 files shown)


laurenzlong (changes requested):
@dionlow I'm hesitant to modify the flow since it adds additional friction and many consumers will likely see the first checkmark and think that they're done and then leave the page. 

The backend will delete the connection if there are no installations that use it anymore. So I think the fix is to ensure that the connections list is refreshed when an installation is deleted, then it won't auto-trigger a reinstall. Feel free to start a new PR for that.

dionlow:
@laurenzlong - is it possible that another proxy integration will still use the connection or are connections 1-1? i.e. what if I have 2 hubspot proxy integrations or 1 salesforce, 1 hubspot?

laurenzlong:
@dionlow it is technically possible, but I think highly unlikely that a builder will set things up that way, so I'm ok with that being an edge case we don't handle well.

dionlow:
Closing PR and making a new one: https://github.com/amp-labs/react/pull/788

PR #813, opened 2025-01-16
[ENG-1805] Bug: field mapping should not persist old mappings

jlimatampersand:
This PR fixes the field mapping bug where the field mapToName fed to `<InstallIntegration>` component changes, bug was introduced because previously selected mappings were not removed even thought it is removed from UI. 

This PR removes deprecated mappings that was present prior to code change to builders. 

## Test Results
- Before Update
Screenshot 2025-01-16 at 3 36 54 PM

- After Code Update and updating installation
Screenshot 2025-01-16 at 3 37 00 PM

### DIFF
--- a/src/components/Configure/content/fields/FieldMappings/RequiredFieldMappings.tsx
+++ b/src/components/Configure/content/fields/FieldMappings/RequiredFieldMappings.tsx
@@ -18,6 +18,26 @@ export function RequiredFieldMappings() {
   const { fieldMapping } = useInstallIntegrationProps();
   const { isError, removeError } = useErrorState();
 
+  const onSelectChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    const { value, name } = e.target;
+    if (!value) {
+      // if place holder value is chosen, we don't change state
+      return;
+    }
+
+    if (selectedObjectName) {
+      setFieldMapping(selectedObjectName, setConfigureState, [{
+        field: name,
+        value,
+        idDeleted: false,
+      }]);
+    }
+
+    if (isError(ErrorBoundary.MAPPING, name)) {
+      removeError(ErrorBoundary.MAPPING, name);
+    }
+  };
+
   const integrationFieldMappings = useMemo(() => {
     // 1. Extract required map fields from configureState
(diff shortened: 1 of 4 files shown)


laurenzlong (changes requested):
as discussed offline:
- used "outdatedKeys" not "deprecatedKeys"
- modify the signature of setFieldMappingReducer have the newValue param be `string | null`, so you don't have to pass in `isDeleted` everywhere.

src/components/Configure/content/fields/FieldMappings/RequiredFieldMappings.tsx
```diff
@@ -53,6 +58,21 @@ export function RequiredFieldMappings() {
...
   }, [configureState, fieldMapping, selectedObjectName]);
 
+  const selectedFieldMappings = configureState?.read?.selectedFieldMappings || {};
+  const selectedKeys = Object.keys(selectedFieldMappings);
+  const allowedKeys = integrationFieldMappings.map((field) => field.mapToName);
+
+  const outdatedKeys = findOutdatedKeys(selectedKeys, allowedKeys);
+
+  if (!!selectedObjectName && outdatedKeys.length) {
+    setFieldMapping(selectedObjectName, setConfigureState, outdatedKeys.map((key) => ({
```
laurenzlong, line 68: can you add a quick comment to explain, something like:

```
For old field mappings that have now been removed by the builder, unset the values for those keys.
```

laurenzlong (approved):

src/components/Configure/content/fields/FieldMappings/setFieldMapping.ts
```diff
@@ -3,13 +3,26 @@ import { Draft } from 'immer';
...
+export type MappingFields = {
+  field: string
+  value: string | null
+};
+
 function setFieldMappingProducer(
   draft: Draft<ConfigureState>,
-  fieldKey: string,
-  newValue: string,
+  fields: Array<MappingFields>,
```
dionlow, line 13: @jlimatampersand does this field need to be an array? setFieldMappingProducer originally set a single field, value.

PR #815, opened 2025-01-17
[ENG-1784] feat: add state management for value default write

dionlow:
### Summary
- add "add" ability for default value write field state
- add "delete" ability for default value write field state
- adds setValueDefaultWriteField producer to modify state (using immer)
- supports showing/hiding default value mapping section only when write access is selected.
- save / modify functionality 

note: This feature is still behind feature flag `DEFAULT_VALUE_FF `. You will need to manually set this in the code to turn this on.

#### followup
Add display name support to fields for default value (needs to fetch from hydrated revision)

#### demo
default-value-state-managment

#### updated ui based on feedback

Screenshot 2025-01-16 at 5 32 24 PM

### DIFF
--- a/src/components/Configure/content/fields/FieldDefaultValueMapping/FieldDefaultValueTable.tsx
+++ b/src/components/Configure/content/fields/FieldDefaultValueMapping/FieldDefaultValueTable.tsx
@@ -46,12 +46,6 @@ export function FieldDefaultValueTable({
         paddingBottom: '1rem', display: 'flex', flexDirection: 'column', gap: '.5rem',
       }}
       >
-        {defaultValueList.length === 0 && <div>No default values</div>}
-        {defaultValueList?.length > 0 && (
-        <div style={{ height: '1rem' }}>
-          <div>Field Default Values</div>
-        </div>
-        )}
         {defaultValueList.map(({ field, fieldDisplayName, defaultValue: df }) => (
           <div
             key={`${field}-${df}`}
@@ -70,6 +64,13 @@ export function FieldDefaultValueTable({
         ))}
       </div>
       <NewDefaultValueUI objectName={objectName} onAddDefaultValue={onAddDefaultValue} />
+      <div style={{
+        textAlign: 'right',
+        padding: '.25rem 0',
+        color: 'var(--amp-colors-text-muted)',
+      }}
+      >click + to confirm
+      </div>
     </>
   );
 }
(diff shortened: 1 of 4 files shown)


src/components/Configure/content/fields/FieldDefaultValueMapping/FieldDefaultValueTable.tsx
```diff
@@ -18,32 +21,31 @@ type FieldDefaultValue = {
...
+  }, [objectName, selectedObjectName, setConfigureState]);
 
   const onDeleteDefaultValue = (field: string) => {
-    setDefaultValueList((prev) => prev.filter((df) => df.field !== field));
+    setValueDefaultWriteField(selectedObjectName || '', objectName, field, null, setConfigureState);
   };
 
+  const defaultValueList: FieldDefaultValue[] = Object.keys(selectedValueDefaultsMap).map((field) => ({
+    field,
+    defaultValue: selectedValueDefaultsMap[field],
```
RajatPawar, line 39: Minor, but this lookup could be in downcase to avoid casing issues (the BE also does this)

src/components/Configure/content/fields/FieldDefaultValueMapping/setValueDefaultWriteField.tsx
```diff
@@ -0,0 +1,70 @@
...
+      const savedWriteObjects = draft.write.savedConfig.selectedWriteObjects;
+      const updatedWriteObjects = draftSelectedWriteFields;
+      const isModified = !isWriteObjectsEqual(savedWriteObjects, updatedWriteObjects);
+      // immer syntax to set a value
+      // eslint-disable-next-line no-param-reassign
+      draft.write.isWriteModified = isModified;
+    }
+
+    // DEBUG: print out the draft
+    // console.debug('Set default value', JSON.stringify(draft?.write, null, 2));
```
RajatPawar, line 50: Don't know if you wanted this in the change

RajatPawar (approved):

RajatPawar (approved):
LGTM!

src/components/Configure/content/fields/FieldDefaultValueMapping/FieldDefaultValueTable.tsx
dionlow: field is guaranteed to match since it's a key from the map in line 37. In this case the typescript type should also guard.

src/components/Configure/content/fields/FieldDefaultValueMapping/setValueDefaultWriteField.tsx
dionlow: I left this in, since debugging with draft state is a little confusing.

PR #820, opened 2025-01-17
[ENG-1784] feat: adds displayName lookup to field default value list

dionlow:
### Summary 
This PR adds the default value displayname looked up via the hydratedRevision.
- adds readAction into hydratedRevision context

#### screenshot
Displayname "Account Currency" is pulled correctly.
Screenshot 2025-01-17 at 10 04 42 AM

### DIFF
--- a/src/components/Configure/content/fields/FieldDefaultValueMapping/FieldDefaultValueTable.tsx
+++ b/src/components/Configure/content/fields/FieldDefaultValueMapping/FieldDefaultValueTable.tsx
@@ -1,5 +1,8 @@
 import { useCallback, useMemo } from 'react';
 
+import { HydratedIntegrationFieldExistent, HydratedIntegrationRead } from 'services/api';
+import { useHydratedRevision } from 'src/components/Configure/state/HydratedRevisionContext';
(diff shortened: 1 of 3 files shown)


src/components/Configure/content/fields/FieldDefaultValueMapping/FieldDefaultValueTable.tsx
```diff
@@ -18,9 +21,25 @@ type FieldDefaultValue = {
...
+ * gets field displayName from hydrated revision, object.allFields
+ * fallback to field name if not found
+ *  */
+const getDisplayNameFromField = (
+  field: string,
+  objectName: string,
+  readAction: HydratedIntegrationRead,
+) => {
+  const object = readAction && getObjectFromAction(readAction, objectName);
+  const allFields = object?.allFields as HydratedIntegrationFieldExistent[] || [];
```
laurenzlong, line 35: we now have `allFieldsMetadata` which is a map already, not a list so  you should use that and then you don't need to call `find` in the next line.

laurenzlong (approved):

PR #825, opened 2025-01-18
feat: show value mapping errors on ui

jatinsandilya:
**Demo**: https://www.loom.com/share/6f1b82af850b4cff94f2b5d515077df8


src/components/Configure/content/fields/ValueMapping/ValuesMapping.tsx
```diff
@@ -129,14 +140,23 @@ export function ValueMappings() {
...
-                    fieldName={field?.fieldName || ''}
-                  />
+                  <>
+                    <ValueMappingItem
+                      key={`${value.mappedValue}-${field.fieldName}`}
+                      allValueOptions={configureState?.read?.allFieldsMetadata?.[field.fieldName!]?.values || []}
+                      mappedValue={value}
+                      onSelectChange={onSelectChange}
+                      fieldName={field?.fieldName || ''}
+                      hasError={!!errors[field.fieldName!]?.[value.mappedValue as string]}
```
dionlow, line 150: are we able to avoid the type cast here?

dionlow (changes requested):
added some comments on using immer for keeping objects immutable

src/components/Configure/content/fields/ValueMapping/ValueMappingItem.tsx
```diff
@@ -62,10 +66,31 @@ export function ValueMappingItem(
 
         if (Object.values(selectedValueMappingForField)
           .some((mapping) => mapping === item.value && mapping !== fieldValue)) {
-          console.error(`Each ${fieldName} must be mapped to a unique value`);
+          let duplicateKeys = Object.entries(selectedValueMappingForField)
+            .filter(([_, mapping]) => mapping === item.value && mapping !== fieldValue)
+            .map(([key]) => key);
+          duplicateKeys = [...duplicateKeys, mappedValue.mappedValue];
+          setErrors((prev: any) => ({
+            ...prev,
```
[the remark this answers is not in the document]
jatinsandilya: @dionlow I ended up refactoring this section to use re-use the ErrorContext and `useErrorState` and switched `let` with `const`

src/components/Configure/content/fields/ValueMapping/ValuesMapping.tsx
```diff
@@ -61,8 +61,8 @@ export function ValueMappings() {
         );
       }
 
-      if (isError(ErrorBoundary.MAPPING, name)) {
-        removeError(ErrorBoundary.MAPPING, name);
+      if (isError(ErrorBoundary.VALUE_MAPPING, name)) {
```
jatinsandilya, line 64: Added a separate boundary here, so they can be reset independently

src/components/Configure/content/fields/ValueMapping/ValueMappingItem.tsx
```diff
@@ -43,56 +51,108 @@ export function ValueMappingItem(
...
+      value: f.value,
+    })),
+    [allValueOptions],
+  );
+
+  const onValueChange = useCallback(
+    (item: { value: string } | null) => {
+      if (!item) return;
+
+      if (
```
dionlow, line 75: adding some comments or naming this logic could be good.

src/components/Configure/content/fields/ValueMapping/ValueMappingItem.tsx
```diff
@@ -43,56 +51,108 @@ export function ValueMappingItem(
...
+        Object.values(selectedValueMappingForField).some(
+          (mapping) => mapping === item.value && mapping !== fieldValue,
+        )
+      ) {
+        const duplicateKeys = [
```
dionlow, line 80: nit: Object.entries, filter, and map produces a new array. we likely do not need to ... spread operator to create a new array again.

dionlow (approved):
left some small comments. feel free add address and merge

src/components/Configure/content/fields/ValueMapping/ValueMappingItem.tsx
jatinsandilya: Yep makes sense, adding comments here.

src/components/Configure/content/fields/ValueMapping/ValueMappingItem.tsx
jatinsandilya: Actually this is an object which needs to be converted into an array of values hence the usage of Object.entries is needed here. I've improved the types so its clearer.

PR #843, opened 2025-01-29
[ENG-1837] refactor: use react-query with hydrated revision

dionlow:
### Summary
migrate hydrated revision to use react-query
- replaces legacy api service
- conditional fetch
- cleans up useEffect

note: pending react-query being added into main. 

#### not included
the new query is nested inside the provider, this might be able to be removed in the future since the hydrated revision is now accessible via the react-query global provider now. 

#### renders as expected
Screenshot 2025-01-29 at 4 03 30 PM

### DIFF
--- a/src/components/Configure/state/HydratedRevisionContext.tsx
+++ b/src/components/Configure/state/HydratedRevisionContext.tsx
@@ -1,15 +1,15 @@
 import React, {
-  createContext, useContext, useEffect, useMemo, useState,
+  createContext, useContext, useEffect, useMemo,
 } from 'react';
+import { useQuery } from '@tanstack/react-query';
 
-import { useApiKey } from 'context/ApiKeyContextProvider';
 import { useConnections } from 'context/ConnectionsContextProvider';
 import {
   ErrorBoundary, useErrorState,
 } from 'context/ErrorContextProvider';
 import { useInstallIntegrationProps } from 'context/InstallIIntegrationContextProvider/InstallIntegrationContextProvider';
 import {
-  api, HydratedIntegrationRead, HydratedIntegrationWriteObject, HydratedRevision,
+  HydratedIntegrationRead, HydratedIntegrationWriteObject, HydratedRevision, useAPI,
 } from 'services/api';
 import { ComponentContainerError, ComponentContainerLoading } from 'src/components/Configure/ComponentContainer';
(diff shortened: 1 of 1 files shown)


src/components/Configure/state/HydratedRevisionContext.tsx
```diff
@@ -39,6 +39,37 @@ export const useHydratedRevision = () => {
...
+) => {
+  const getAPI = useAPI();
+  const { selectedConnection } = useConnections();
+  const { integrationId, integrationObj } = useInstallIntegrationProps();
+
+  const connectionId = selectedConnection?.id;
+  const revisionId = integrationObj?.latestRevision?.id;
+
+  return useQuery({
+    queryKey: ['amp', 'hydratedRevision', projectIdOrName, integrationId, revisionId],
```
dionlow, line 54: adds amp prefix to prevent possible query key clash with builder app

src/components/Configure/state/HydratedRevisionContext.tsx
```diff
@@ -39,6 +39,37 @@ export const useHydratedRevision = () => {
...
+
+      const api = await getAPI();
+      return api.revisionApi.getHydratedRevision({
+        projectIdOrName,
+        integrationId,
+        revisionId,
+        connectionId,
+      });
+    },
+    enabled: !!projectIdOrName && !!integrationId && !!revisionId && !!connectionId,
```
dionlow, line 69: does not fetch without these being true.

src/components/Configure/state/HydratedRevisionContext.tsx
```diff
@@ -39,6 +39,37 @@ export const useHydratedRevision = () => {
...
+
+  return useQuery({
+    queryKey: ['amp', 'hydratedRevision', projectIdOrName, integrationId, revisionId],
+    queryFn: async () => {
+      if (!projectIdOrName) throw new Error('projectIdOrName is required');
+      if (!integrationId) throw new Error('integrationId is required');
+      if (!revisionId) throw new Error('revisionId is required');
+      if (!connectionId) throw new Error('connectionId is required');
+
+      const api = await getAPI();
```
dionlow, line 61: uses new api service.

RajatPawar (approved):

PR #845, opened 2025-01-30
chore: removes spacing from value mappings

dionlow:
### Summary
cleans up some spacing, while debugging. 
- ~fix render 0 when no optional field mappings are present~ separated to reduce merge conflicts
- refactors/cleans up value mapping white space.

### DIFF
--- a/src/components/Configure/content/fields/FieldMappings/OptionalFieldMappings.tsx
+++ b/src/components/Configure/content/fields/FieldMappings/OptionalFieldMappings.tsx
@@ -25,12 +25,7 @@ export function OptionalFieldMappings() {
     }
 
     if (selectedObjectName) {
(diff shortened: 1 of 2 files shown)


src/components/Configure/content/fields/FieldMappings/OptionalFieldMappings.tsx
```diff
@@ -77,7 +75,7 @@ export function OptionalFieldMappings() {
             />
           </FormControl>
         ))}
-        {dynamicFieldMappings?.length && (
```
dionlow, line 80: this line was rendering 0.

jatinsandilya: ah great catch!

PR #847, opened 2025-01-30
[ENG-1778, ENG-1837] refactor: connections to use react query (connections part 1) 

dionlow:
### Summary
We add react query to simplify how to manage syncing connections. This is prep work for [ENG-1778] bug. Simplifying logic in connections manager via react query helps avoid re-render issues when switching connections

- simplify connections context
- derive connection from connections list
- remove manual set connections/ connection
- adds loading state to hydrated revision which previously showed an error when connection was set.

#### Previously
1. useEffect, deps, prop changes, and actions from user (auth) trigger a re-fetch of listConnections
2. based on oauth and connections, we use a useEffect connections manager to set the selected connection and keep them in sync

issues: syncing selectedConnection and connections can create a endless cycle has they trigger multiple useEffects to update each other

#### Now
1. we create a useConnectionsListQuery that will only fetch and cache the connections list.
2. instead of tracking a separate state for selected connection, connection is always derived from connection list (the first connection)
3. any mutation like an auth flow login will invalidate connections cache - asking react query to refetch.

we have some placeholder functions like setConnections and setSelectedConnections which will just invalidate the cache behind the scenes until mutate functions with their own cache invalidation calls are also migrated to react-query.

#### followup
part 2: https://github.com/amp-labs/react/pull/848
Oauth connections logic is not fully migrated yet but will still use the connections from react query. All manual connection sets will invalidate cache until deleted in the future.

#### demo
connections-demo

### DIFF
--- a/src/context/ConnectionsContextProvider.tsx
+++ b/src/context/ConnectionsContextProvider.tsx
@@ -1,10 +1,10 @@
 import {
-  createContext, useContext, useEffect, useMemo, useState,
+  createContext, useCallback, useContext, useEffect, useMemo,
 } from 'react';
+import { useQuery, useQueryClient } from '@tanstack/react-query';
 
 import { Connection, useAPI } from 'services/api';
 import { ComponentContainerError, ComponentContainerLoading } from 'src/components/Configure/ComponentContainer';
-import { useIsInstallationDeleted } from 'src/hooks/useIsInstallationDeleted';
 import { handleServerError } from 'src/utils/handleServerError';
 
 import { ErrorBoundary, useErrorState } from './ErrorContextProvider';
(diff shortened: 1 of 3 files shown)


src/context/ConnectionsContextProvider.tsx
```diff
@@ -41,111 +37,85 @@ export const useConnections = (): ConnectionsContextValue => {
...
 };
 
+const useConnectionsListQuery = () => {
+  const { projectIdOrName } = useProject();
+  const { groupRef, provider } = useInstallIntegrationProps();
+  const getAPI = useAPI();
+  return useQuery({
+    queryKey: ['amp', 'connections', projectIdOrName, groupRef, provider],
+    queryFn: async () => {
+      if (!projectIdOrName) throw new Error('Project ID not found.');
```
laurenzlong, line 47:
```suggestion
      if (!projectIdOrName) throw new Error('Project ID or name not found. Please wrap this component inside of AmpersandProvider');
```

laurenzlong (approved):

PR #862, opened 2025-02-06
[ENG-1910] fix: useIsIntegrationInstalled 

dionlow:
### Summary
the `useIsIntegrationInstalled` hook is out of sync from the installations query. After refactoring list installations to use react query, 
- add in some cache invalidation logic to enable the hook state to be synced. 
- fixes some unnecessary fetches

followup to https://github.com/amp-labs/react/pull/844

##### configure
integration installed hook tested in mailmonkey
fix-integration-installed-hook

##### proxy
proxy-test

### DIFF
--- a/src/components/Configure/layout/ProtectedConnectionLayout.tsx
+++ b/src/components/Configure/layout/ProtectedConnectionLayout.tsx
@@ -1,4 +1,5 @@
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
+import { useQueryClient } from '@tanstack/react-query';
 
 import { ApiKeyAuthFlow } from 'components/auth/ApiKeyAuth/ApiKeyAuthFlow';
 import { BasicAuthFlow } from 'components/auth/BasicAuth/BasicAuthFlow';
@@ -54,6 +55,7 @@ export function ProtectedConnectionLayout({
   const { provider: providerFromProps, isIntegrationDeleted } = useInstallIntegrationProps();
   const { selectedConnection, setSelectedConnection } = useConnections();
   useConnectionHandler({ onSuccess });
+  const queryClient = useQueryClient();
 
   const selectedProvider = provider || providerFromProps;
 
@@ -68,6 +70,11 @@ export function ProtectedConnectionLayout({
     });
   }, [apiKey, selectedProvider]);
 
+  const reinstallIntegration = useCallback(() => {
+    queryClient.clear(); // clears all queries in react-query cache
+    resetComponent();
+  }, [resetComponent, queryClient]);
+
   if (!provider && !providerFromProps) {
     throw new Error('ProtectedConnectionLayout must be given a provider prop or be used within InstallIntegrationProvider');
   }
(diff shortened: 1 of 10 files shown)


src/components/Configure/layout/ProtectedConnectionLayout.tsx
```diff
@@ -68,6 +70,11 @@ export function ProtectedConnectionLayout({
     });
   }, [apiKey, selectedProvider]);
 
+  const reinstallIntegration = useCallback(() => {
+    queryClient.clear(); // clears all queries in react-query cache
```
dionlow, line 74: we need to clear cache or will try to retrigger hydrated revision and other calls using an old selected connection.

src/context/InstallIIntegrationContextProvider/InstallIntegrationContextProvider.tsx
```diff
@@ -95,15 +95,17 @@ export function InstallIntegrationProvider({
...
 
   const installation = installations?.[0] || null; // there should only be one installation for mvp
 
   // resets the isIntegrationDeleted state when InstallIntegrationProps change
   useEffect(() => {
     resetIntegrationDeleted();
-  }, [integration, consumerRef, consumerName, groupRef, groupName,
-    onInstallSuccess, onUpdateSuccess, onUninstallSuccess, fieldMapping, resetIntegrationDeleted]);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [integration, consumerRef, consumerName, groupRef, groupName, fieldMapping,
```
dionlow, line 106: commented out dep array that is causing unneccessary re-renders when deleted state is set

src/components/Configure/state/HydratedRevisionContext.tsx
```diff
@@ -38,15 +39,13 @@ export const useHydratedRevision = () => {
...
-) => {
+const useHydratedRevisionQuery = () => {
   const getAPI = useAPI();
-  const { selectedConnection } = useConnections();
+  const { data: connectionsData, isLoading: isConnectionsLoading } = useConnectionsListQuery();
+  const { projectIdOrName } = useProject();
   const { integrationId, integrationObj } = useInstallIntegrationProps();
 
-  const connectionId = selectedConnection?.id;
+  const connectionId = connectionsData?.[0]?.id;
```
laurenzlong, line 48: can be follow up, but it would be great to have a hook called `useConnections` that calls `useConnectionsListQuery` and returns both the connections list, a loading boolean, and a selected connection. So that in the future when it is no longer true that connections is a list of one, we can easily change the logic in one place. Otherwise, there will be a bunch of `connectionsData?.[0]` all over the codebase that will need to be updated.

laurenzlong (approved):

src/components/Configure/state/HydratedRevisionContext.tsx
dionlow: sounds good. I was originally thinking that we'll remove theConnections provider entirely, but we can keep it there to manage future connections.

PR #873, opened 2025-02-10
fix: add error handling try/catch

dionlow:
### Summary
Errors are not able to be read from the response more than once. 
1. We create a clone that will attempt to read instead of creating an unhandled error.
2. Surround with try/catch to prevent unhandled errors

### DIFF
--- a/src/utils/handleServerError.ts
+++ b/src/utils/handleServerError.ts
@@ -18,7 +18,9 @@ export const handleServerError = async (error: any, setError?: (error: string) =
     }
 
     try {
-      const errorBody = await error.response.json(); // If the response body contains error details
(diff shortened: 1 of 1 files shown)


src/utils/handleServerError.ts
```diff
@@ -16,13 +16,20 @@ export const handleServerError = async (error: any, setError?: (error: string) =
     } else {
       console.error(`Error (${status}):`, statusText);
     }
-    const errorBody = await error.response.json(); // If the response body contains error details
```
dionlow, line 19: unhandled error if response has already been read.

eberle1080 (approved):

PR #878, opened 2025-02-12
[ENG-1920] fix: provider display name

dionlow:
### Summary
Sometimes provider displayName does not show correctly. We move the util function inside a new hook `useProvider` which extends the providerInfoQuery with derived state.
- adds useProvider hook
- enforce providerInfo as when getting the displayname

#### Screenshot
Screenshot 2025-02-12 at 2 03 57 PM

Screenshot 2025-02-12 at 2 03 33 PM

### DIFF
--- a/src/components/Configure/layout/ProtectedConnectionLayout.tsx
+++ b/src/components/Configure/layout/ProtectedConnectionLayout.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useEffect } from 'react';
-import { useQuery, useQueryClient } from '@tanstack/react-query';
+import { useQueryClient } from '@tanstack/react-query';
 
 import { ApiKeyAuthFlow } from 'components/auth/ApiKeyAuth/ApiKeyAuthFlow';
 import { BasicAuthFlow } from 'components/auth/BasicAuth/BasicAuthFlow';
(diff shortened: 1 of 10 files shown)


src/components/Configure/layout/ProtectedConnectionLayout.tsx
```diff
@@ -30,40 +30,19 @@ interface ProtectedConnectionLayoutProps {
...
-
 export function ProtectedConnectionLayout({
   provider, consumerRef, consumerName, groupRef, groupName, children, onSuccess, onDisconnectSuccess,
   resetComponent,
 }: ProtectedConnectionLayoutProps) {
   const {
     data: providerInfo, isLoading: isProviderLoading, isError, error: providerInfoError,
-  } = useProviderInfo(provider);
+    providerName, selectedProvider,
+  } = useProvider(provider);
```
dionlow, line 40: note passing in provider is required for components that are also used under Connect component and do not have access to provider from Installations (InstallIntegration component)

src/hooks/useProvider.tsx
```diff
@@ -0,0 +1,46 @@
+import { useQuery } from '@tanstack/react-query';
+
+import { useAPI } from 'services/api';
+import { useInstallIntegrationProps } from
+  'src/context/InstallIIntegrationContextProvider/InstallIntegrationContextProvider';
+import { getProviderName } from 'src/utils';
+
+const useProviderInfoQuery = (provider?: string) => {
```
dionlow, line 8: this query is just moved and renamed with Query.

jatinsandilya (approved):

PR #880, opened 2025-02-13
fix: FieldMapping type export

laurenzlong:
Looks like the export broke at some point, and we were accidentally exporting the React component named FieldMapping and not the type. This PR rename the React component to FieldMappingRow to differentiate it from the FieldMapping type (which is a prop of InstallIntegration), and also updated the docstrings to make them more builder-friendly.


src/components/Configure/InstallIntegration.tsx
```diff
@@ -20,19 +20,35 @@ export interface MappedValue {
   mappedDisplayValue: string;
 }
 
-export type FieldMappingWithMappedValues = IntegrationFieldMapping & {
```
laurenzlong, line 23: I had to redeclare the fields of IntegrationFieldMapping since otherwise the exported type didn't show all the fields (unless we also exported IntegrationFieldMapping, which I don't want to do since it would be confusing)

dionlow (approved):

PR #927, opened 2025-03-10
build: update handle error to look at details and remedy

dionlow:
### Summary
updates handle error to add error details and body, previously only looked at causes array which turn up empty sometimes.

#### demo
Screenshot 2025-03-10 at 3 30 50 PM

### DIFF
--- a/src/utils/handleServerError.ts
+++ b/src/utils/handleServerError.ts
@@ -24,8 +24,14 @@ export const handleServerError = async (error: any, setError?: (error: string) =
 
       // https://github.com/amp-labs/openapi/blob/3bc3ab75c3071763e1117f697be3e0fcb636972c/problem/problem.yaml#L85
       // All errors returned by the Ampersand API conform to this format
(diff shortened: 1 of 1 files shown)


src/utils/handleServerError.ts
```diff
@@ -24,8 +24,14 @@ export const handleServerError = async (error: any, setError?: (error: string) =
...
-      console.error('[Error Message]', errorMsg);
+      const errorMsg = errorBody?.causes?.join('\n') || errorBody?.detail;
+      if (!errorMsg) {
+        console.error('Unexpected error response:', errorBody);
+      } else {
+        console.error('[Error Message]', errorMsg);
+        if (errorBody?.remedy) { console.error('[Remedy]', errorBody.remedy); }
+      }
+
       if (setError) setError(errorMsg);
```
laurenzlong, line 35: can you also add the remedy to this errMsg string? so we can display the remedy in the UI

laurenzlong (approved):

PR #936, opened 2025-03-14
[ENG-1980] feat: add hydrated revision error ui, disconnect button + error

dionlow:
### Summary 
Hydrated error state console logs error, but does not add detail to UI error. 
- adds UI causes error for hydrated revision
- adds disconnect button 
- adds disconnect error UI 

#### demo
note demo does not show causes arrary in UI since hydrated revision error is triggered by react-query error; does not recreate hydrated revision error from our amperand server endpoint. similiar logic is used in the disconnect button however.

hydrated-revision-error-state

### DIFF
--- a/src/components/Configure/InstallIntegration.tsx
+++ b/src/components/Configure/InstallIntegration.tsx
@@ -137,7 +137,7 @@ export function InstallIntegration(
             groupName={groupName}
             resetComponent={reset}
           >
-            <HydratedRevisionProvider>
+            <HydratedRevisionProvider resetComponent={reset}>
               <ConditionalProxyLayout>
                 <ConfigurationProvider>
                   <ObjectManagementNav>
--- a/src/components/Configure/state/HydratedRevisionContext.tsx
+++ b/src/components/Configure/state/HydratedRevisionContext.tsx
@@ -1,5 +1,5 @@
 import React, {
-  createContext, useContext, useEffect, useMemo,
+  createContext, useContext, useEffect, useMemo, useState,
 } from 'react';
 import { useQuery, useQueryClient } from '@tanstack/react-query';
 
@@ -87,6 +87,7 @@ export function HydratedRevisionProvider({
   const { integrationId, integrationObj } = useInstallIntegrationProps();
   const { isError, removeError, setError } = useErrorState();
   const errorIntegrationIdentifier = integrationObj?.name || integrationId;
+  const [readeableErrorMsg, setReadableErrorMsg] = useState<string | null>(null);
 
   const {
     data: hydratedRevision,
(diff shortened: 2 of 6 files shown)


src/components/Configure/InstallIntegration.tsx
```diff
@@ -137,7 +137,7 @@ export function InstallIntegration(
             groupName={groupName}
             resetComponent={reset}
           >
-            <HydratedRevisionProvider>
+            <HydratedRevisionProvider resetComponent={reset}>
```
dionlow, line 140: reset is untested.

src/components/Connect/RemoveConnectionButton.tsx
```diff
@@ -68,6 +69,9 @@ export function RemoveConnectionButton({
             onDisconnectSuccess?.(selectedConnection);
             resetComponent(); // reset / refresh the Connect Provider component
           },
+          onError: (error) => {
```
dionlow, line 72: we move the handleServerError in the second error handling step so we can pass in onDisconnectError.

dionlow:
Screenshot 2025-03-14 at 3 42 06 PM

just a note that copilot suggested correcting the variable name spelling (good). but then broke the build since it did not change all variables except the one instantiating.

src/components/Configure/state/HydratedRevisionContext.tsx
```diff
@@ -117,9 +125,25 @@ export function HydratedRevisionProvider({
   if (isError(ErrorBoundary.HYDRATED_REVISION, errorIntegrationIdentifier)) {
     const intNameOrId = integrationObj?.name || integrationId || 'unknown integration';
     const errorMsg = `Error retrieving integration details for '${intNameOrId
-    }. This is sometimes caused by insufficient permissions with your credentials'`;
-
-    return <ComponentContainerError message={errorMsg} />;
+    }. This is sometimes caused by insufficient permissions with your credentials. ' 
+    ${readableErrorMsg ? `: ${readableErrorMsg}` : ''}`;
```
laurenzlong, line 129: Let's get rid of "This is sometimes caused by insufficient permissions with your credentials." that was a generic message, and was causing a lot of confusion, since it's usually not actually a permission issue.

src/components/Configure/state/HydratedRevisionContext.tsx
```diff
@@ -117,9 +125,25 @@ export function HydratedRevisionProvider({
...
+
+    return (
+      <ComponentContainerError message={errorMsg}>
+        <div style={{
+          display: 'flex', flexDirection: 'column', gap: '1rem', paddingTop: '1rem',
+        }}
+        >
+          {connectionError && <InnerErrorTextBox message={connectionError} />}
+          <RemoveConnectionButton
+            buttonText="Remove Connection"
```
laurenzlong, line 139: this text should say "Try again".  "remove connection" is confusing, since even though we have a connection created in the backend at this point, the user doesn't know that and from their perspective the connection wasn't successful yet.

src/utils/handleServerError.ts
```diff
@@ -32,8 +32,10 @@ export const handleServerError = async (error: any, setError?: (error: string) =
         if (errorBody?.remedy) { console.error('[Remedy]', errorBody.remedy); }
       }
 
+      const combinedErrorMessage = `[Error Message] ${errorMsg}
```
laurenzlong, line 35: Not sure `Error Message` is actually necessary. Since this text is rendered in a red box, it's pretty obvious it's an error message.

laurenzlong (approved):

src/utils/handleServerError.ts
dionlow: This logic happens in the logging and I reuse the same logic for UI. Would you wan this to removed for both?

PR #948, opened 2025-03-19
[ENG-2033] feat: add includeEntitlements to getProject

dionlow:
### Summary
adds entitlements 

#### screenshot
Screenshot 2025-03-19 at 4 53 23 PM
Screenshot 2025-03-19 at 4 53 14 PM

### DIFF
--- a/src/context/ProjectContextProvider.tsx
+++ b/src/context/ProjectContextProvider.tsx
@@ -17,7 +17,7 @@ const useProjectQuery = (projectIdOrName: string) => {
     queryKey: ['project', projectIdOrName],
     queryFn: async () => {
       const api = await getAPI();
-      return api.projectApi.getProject({ projectIdOrName });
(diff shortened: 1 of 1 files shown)


laurenzlong (approved):

RajatPawar:
@dionlow This might already be the case, but is it possible to request entitlements only when figuring out the footer, and not everywhere we need the project context? Asking for entitlements is an extra call to our billing provider, and even though we cache the response after the first call, it might add latency wherever this is being used

dionlow:
@RajatPawar - is latency or cost very considerable to add entitlements? The other option is to create a second query with entitlements, but the we're possibly fetching project twice with two separate caches. Alternatively if there is an entitlements endpoint, that we pass in projectId we leave the caches separate regardless. 

If performance is not super big, then we keep this for simplicity.

dionlow:
Merging as the followup is primarily performance

PR #949, opened 2025-03-20
performance: useProjectWithEntitlementsQuery

dionlow:
### Summary 
Creates a separate query for fetching projects with entitlements. 

Trade offs:
- fetches projects with entitlements only when Ampersand footer is rendered 
- double fetches project whenever footer is present 

notes: if performance is a concern than, we should consider entitlements as a separate endpoint.


dionlow:
let me know your thoughts @RajatPawar

RajatPawar (approved):

dionlow:
noted that the second query takes an additional second.

PR #951, opened 2025-03-20
feat: show installed success box for proxy-only & subscribe-only 

dionlow:
### Summary
See that we show an integration success flow when read and write are not available
- should work for both proxy and subscribe-only integrations.

#### Merged into branch
followup to supporting proxy and subscribe create installation
- move create installation to a react-query mutation hook
- add support for subscribe-only installations
- refactor proxy-only installations to use the new mutation hook

#### error msg
Screenshot 2025-03-21 at 4 27 56 PM

#### proxy enabled
Screenshot 2025-03-21 at 4 33 58 PM

#### proxy not enabled
Screenshot 2025-03-21 at 4 09 17 PM

#### tested proxy and subscribe
subscribe-only


laurenzlong (changes requested):
This is actually now always sending `proxy: {enabled: true}` even if the integration didn't have proxy actions. Can you ensure that you only send that if the revision actually has proxy actions?

otherwise, you can send an empty object as the config content. (Please test this, I'm pretty sure server shouldn't complain about it based on reading code, but need to verify). 

also it turns out that subscribe actions does require read actions currently, so this is no longer blocking launch.

dionlow:
Updated branch to have support for proxy enabled and not-enabled upon creation. added more error handling, and migrated to react query.

src/components/Configure/layout/ConditionalHasConfigurationLayout/ConditionalHasConfigurationLayout.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const { read, write } = hydratedRevision?.content ?? {};
+  return (!read && !write);
+};
+
+interface ConditionalHasConfigurationLayoutProps {
+  children: React.ReactNode;
+}
+
+/**
+ * if the hydratedRevision only has proxy actions (or no configuration required),
```
laurenzlong, line 27: Comment outdated

laurenzlong (approved):

laurenzlong:
Can you revise the error message to say "Installation already exists for groupRef XYZ" or something like that. "Conflicting installation" is confusing

PR #960, opened 2025-03-28
[ENG-2099] feat: [headless] adds InstallationProvider 

dionlow:
### Summary 
Adds a provider for shared params to headless lib. 
- creates a new folder `headless`
- adds InstallationProvider

note: integrationId is specified rather than integration. I find it difficult to distinguish integrationId, integration(object), and integrationName. I change this to integrationId so we're not fetching the object to get the id. spec change may be required.

Consider renaming to HeadlessInstallationProvider.

### DIFF
--- a/src/headless/InstallationProvider.tsx
+++ b/src/headless/InstallationProvider.tsx
@@ -1,0 +1,61 @@
+// headless installation provider
+
+import {
+  createContext,
+  useContext,
+  useMemo,
+} from 'react';
+
+// Define the context value type
+interface InstallationContextValue {
+  integrationId: string;
+  consumerRef: string;
+  consumerName?: string;
+  groupRef: string;
+  groupName?: string;
+}
+// Create a context to pass down the props
+const InstallationContext = createContext<InstallationContextValue>({
+  integrationId: '',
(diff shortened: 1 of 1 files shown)


src/headless/InstallationProvider.tsx
```diff
@@ -0,0 +1,61 @@
...
+  consumerRef: string,
+  consumerName?: string,
+  groupRef: string,
+  groupName?: string,
+  children: React.ReactNode,
+}
+
+// Wrap your parent component with the context provider
+export function InstallationProvider({
+  children, integrationId, consumerRef, consumerName, groupRef, groupName,
```
laurenzlong, line 46: can we actually require integrationName instead of integrationId? it is more user-friendly (since if someone has multiple Ampersand projects for their different environments, the integration IDs will be different for each Ampersand project but the integration name would be the same)

laurenzlong (approved):

src/headless/InstallationProvider.tsx
dionlow: I'll create a ticket to track the integrationName and integrationId support.

PR #977, opened 2025-04-04
[ENG-2119] feat: add update installation hook

dionlow:
### Summary
adds update installation mutation and headless update installation hook
part 3 of https://linear.app/ampersand/issue/ENG-2106/[headless]-useinstallationsmanager

#### testing
headless folder is not exposed to public. will be tested in future along side other installation hooks


RajatPawar (approved):
also adding @greptileai to test review capability

PR #989, opened 2025-04-08
feat: adds manage tab placeholder

dionlow:
### Summary
Adds ui tab for update connections
- adds navigation to manage tab
- adds manage content placeholder
- adds feature flag to switch out uninstall and manage tabs

#### demo
add-manage-tab

#### ff turned off
Screenshot 2025-04-08 at 4 49 39 PM


dionlow:
@RajatPawar - yeah good call out on the design. I think there was a space put there in the original figma. We may be changing that as we populate the nav a bit more.

PR #990, opened 2025-04-09
[ENG-2147] feat: adds authentication section in manage tab

dionlow:
### Summary
adds authenticate section for the manage tab. 
- shows connection workspace and status. 
- adds salesforce exception for subdomain

Screenshot 2025-04-09 at 12 14 48 PM

Screenshot 2025-04-09 at 12 16 56 PM

### DIFF
--- a/src/components/Configure/content/manage/AuthenticationSection.tsx
+++ b/src/components/Configure/content/manage/AuthenticationSection.tsx
@@ -1,0 +1,32 @@
+import { useInstallIntegrationProps } from
+  'src/context/InstallIIntegrationContextProvider/InstallIntegrationContextProvider';
+import { useConnectionQuery } from 'src/hooks/query';
+
+import { FieldHeader } from '../fields/FieldHeader';
+
+import styles from './authenticate.module.css';
+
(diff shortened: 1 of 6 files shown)


src/components/Configure/content/manage/AuthenticationSection.tsx
```diff
@@ -0,0 +1,32 @@
+import { useInstallIntegrationProps } from
+  'src/context/InstallIIntegrationContextProvider/InstallIntegrationContextProvider';
+import { useConnectionQuery } from 'src/hooks/query';
+
+import { FieldHeader } from '../fields/FieldHeader';
+
+import styles from './authenticate.module.css';
+
+function AuthenticationRow({ label, value }: { label: string; value: string | undefined; }) {
```
RajatPawar, line 9: minor: this seems reusable enough that it could be a row component rather than an authentication row component

dionlow: yeah, but we don't currently need this anywhere else in the design. Will leave in this section and refactor to a general folder when when we do.

PR #1012, opened 2025-04-17
fix: form submission/re-render issue

dionlow:
#### problem: 
clicking submit on the update connection flow was causing the app to reload and re-render. 

#### root cause
This bug is caused by an underlying implementation that nests the borrowed `AuthKeyAuthForm` submit button in the `ConfigureBase` save/create installation form. 

#### solution
This PR does a quick fix and turns off the submit type button, however the bug shows a tech debt issue with the current tab structure.

- make submit button exception
- extra: removes extra calls to connection in the auth tab 

### todo
will need to also fix this in the other auth flows. 

### demo
update-api-key-fixed

### DIFF
--- a/src/components/Configure/content/manage/AuthenticationSection.tsx
+++ b/src/components/Configure/content/manage/AuthenticationSection.tsx
@@ -1,7 +1,4 @@
 import { useConnections } from 'src/context/ConnectionsContextProvider';
-import { useInstallIntegrationProps } from
-  'src/context/InstallIIntegrationContextProvider/InstallIntegrationContextProvider';
-import { useConnectionQuery } from 'src/hooks/query';
 
 import { FieldHeader } from '../fields/FieldHeader';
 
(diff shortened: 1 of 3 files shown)


src/components/Configure/content/manage/AuthenticationSection.tsx
```diff
@@ -16,20 +13,16 @@ function AuthenticationRow({ label, value }: { label: string; value: string | un
   );
 }
 export function AuthenticationSection() {
-  const { installation } = useInstallIntegrationProps();
   const { selectedConnection } = useConnections();
-  const connectionId = installation?.connection?.id || selectedConnection?.id || '';
-  const { data: connection } = useConnectionQuery({ connectionId });
-
-  const isSalesforce = connection?.provider === 'salesforce';
+  const isSalesforce = selectedConnection?.provider === 'salesforce';
```
RajatPawar, line 17: unrelated, but it'd be nice to have these cases in a `utils/provider` sort of a file so that when the catalog is able to handle these cases, we can refactor quicker

RajatPawar (approved):

src/components/Configure/content/manage/AuthenticationSection.tsx
dionlow: this case is an exception, since we have a custom use case just for salesforce. if we see more of this, can def refactor.

PR #1017, opened 2025-04-19
feat: update copy for Manage tab, required fields heading, etc.

laurenzlong:
Updates copy for a few things:
- modify headings to sentence case not title case (please try to keep this in mind in the future @dionlow , we always use sentence case in our website, docs, product, i.e. only first word is capitalized).
- fix the uninstallation copy, we've gotten feedback that the current copy is too scary
- fix a few "bugs" in the usage of names vs display names

Screenshot 2025-04-18 at 5 21 07 PM

### DIFF
--- a/src/components/Configure/content/fields/FieldMappings/FieldMappingRow.tsx
+++ b/src/components/Configure/content/fields/FieldMappings/FieldMappingRow.tsx
@@ -127,7 +127,7 @@ export function FieldMappingRow({
           marginBottom: '.25rem',
         }}
         >
-          <span style={{ fontWeight: 500 }}>{field.mapToDisplayName}</span>
+          <span style={{ fontWeight: 500 }}>{field.mapToDisplayName ?? field.mapToName}</span>
           <span>
             {field?.prompt && <LabelTooltip id={`tooltip-id-${field?.prompt}`} tooltipText={field?.prompt} />}
           </span>
--- a/src/components/Configure/content/fields/RequiredFields.tsx
+++ b/src/components/Configure/content/fields/RequiredFields.tsx
@@ -12,7 +12,7 @@ export function RequiredFields() {
 
   return (
     <>
-      <FieldHeader string={`${appName} reads the following ${selectedObjectName} fields`} />
+      <FieldHeader string={`${appName} reads the following fields`} />
       <div style={{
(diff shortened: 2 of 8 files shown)


src/components/Configure/content/fields/FieldMappings/FieldMappingRow.tsx
```diff
@@ -127,7 +127,7 @@ export function FieldMappingRow({
           marginBottom: '.25rem',
         }}
         >
-          <span style={{ fontWeight: 500 }}>{field.mapToDisplayName}</span>
+          <span style={{ fontWeight: 500 }}>{field.mapToDisplayName ?? field.mapToName}</span>
```
laurenzlong, line 130: fixes the bug where if field only has mapToName but not mapToDisplayName (which is a valid thing to do in amp.yaml), then it looks like this:
Screenshot 2025-04-16 at 10 39 23 PM

src/components/Configure/content/fields/RequiredFields.tsx
```diff
@@ -12,7 +12,7 @@ export function RequiredFields() {
 
   return (
     <>
-      <FieldHeader string={`${appName} reads the following ${selectedObjectName} fields`} />
+      <FieldHeader string={`${appName} reads the following fields`} />
```
laurenzlong, line 15: removed it, since we weren't using display name of the object, so it looked weird:
Screenshot 2025-04-18 at 4 23 58 PM

src/components/Configure/content/manage/AuthenticationSection.tsx
```diff
@@ -14,15 +15,17 @@ function AuthenticationRow({ label, value }: { label: string; value: string | un
...
+  const workspaceLabel = `${providerName} ${workspaceString}`;
 
   return (
     <>
-      <FieldHeader string="Authentication" />
+      <FieldHeader string="Connection details" />
       <div style={{ paddingBottom: '1rem' }}>
-        <AuthenticationRow label={workspaceString} value={selectedConnection?.providerWorkspaceRef} />
-        <AuthenticationRow label="Connection Status" value={selectedConnection?.status} />
+        <AuthenticationRow label={workspaceLabel} value={selectedConnection?.providerWorkspaceRef} />
```
dionlow, line 27: just to confirm you want the labels to be lowercase.

dionlow (approved):
LGTM. good catch on the sentence case.

PR #1049, opened 2025-05-01
fix: oauth cached url

dionlow:
### Summary
part 2 followup to https://ampersand-company.slack.com/archives/C082KPRN4R4/p1746117090872019

problem: url can be cached by react-query which will auto popup the window even when not ready
solution: add a separate state to track whether url is active
- adds additional checks for error and oauth url
- fixes async call to refetch oauth url (will await for response, instead of autofire)

#### demo
oauth-fix-cache


src/components/auth/Oauth/AuthorizationCode/WorkspaceEntry/WorkspaceOauthFlow.tsx
```diff
@@ -32,6 +32,8 @@ export function WorkspaceOauthFlow({
 }: WorkspaceOauthFlowProps) {
   const [workspace, setWorkspace] = useState<string>("");
   const [localError, setLocalError] = useState<string | null>(null);
+  // keeps track of whether the OAuth popup URL should be passed to the OAuthWindow
```
laurenzlong, line 35:
```suggestion
  // keeps track of whether the OAuth popup URL should be passed to the OAuthWindow
  // when this is false, then OAuthWindow component will not pop up a window.
```

laurenzlong (approved):

PR #1050, opened 2025-05-02
feat: allow metadata as input

RajatPawar:
Screenshot 2025-05-01 at 5 04 02 PM
Screenshot 2025-05-01 at 5 22 02 PM
Screenshot 2025-05-01 at 5 22 12 PM
Screenshot 2025-05-01 at 5 22 15 PM
Screenshot 2025-05-01 at 5 22 49 PM

### DIFF
--- a/src/components/auth/BasicAuth/BasicAuthContent.tsx
+++ b/src/components/auth/BasicAuth/BasicAuthContent.tsx
@@ -1,5 +1,5 @@
 import { useState } from "react";
-import { ProviderInfo } from "@generated/api/src";
+import { ProviderInfo, MetadataItemInput } from "@generated/api/src";
 import { AuthErrorAlert } from "src/components/auth/AuthErrorAlert/AuthErrorAlert";
 import { FormComponent } from "src/components/form";
 import { Button } from "src/components/ui-base/Button";
@@ -20,6 +20,7 @@ type BasicAuthFormProps = {
   handleSubmit: (form: BasicCreds) => void;
   isButtonDisabled?: boolean;
   buttonVariant?: "ghost";
+  requiredProviderMetadata?: MetadataItemInput[];
 };
 
 export function BasicAuthForm({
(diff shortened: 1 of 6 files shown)


src/components/auth/BasicAuth/BasicAuthFlow.tsx
```diff
@@ -36,6 +36,15 @@ export function BasicAuthFlow({
             username: user,
             password: pass,
           },
+          ...(providerMetadata && {
+            providerMetadata: Object.entries(providerMetadata).reduce((acc, [name, value]) => {
```
laurenzlong, line 40: This technically works, but this would be a lot clearer as as a map function rather than a reduce function.

src/components/auth/BasicAuth/BasicAuthContent.tsx
```diff
@@ -47,7 +53,32 @@ export function BasicAuthForm({
   // TODO(ENG-1424): Uncomment the following line when we handle this properly.
   // const isPassValid = password.length > 0;
   // const isSubmitDisabled = isButtonDisabled || !isUserValid || !isPassValid;
-  const isSubmitDisabled = isButtonDisabled;
+  const onHandleSubmit = () => {
```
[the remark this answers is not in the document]
laurenzlong: +1 but this can be a follow up.

laurenzlong (approved):

src/components/auth/BasicAuth/BasicAuthContent.tsx
```diff
@@ -28,10 +29,15 @@ export function BasicAuthForm({
   handleSubmit,
   isButtonDisabled,
   buttonVariant,
+  requiredProviderMetadata = [],
 }: BasicAuthFormProps) {
   const [show, setShow] = useState(false);
   const onToggleShowHide = () => setShow((prevShow) => !prevShow);
-  const [formData, setFormData] = useState({ username: "", password: "" });
+  const [formData, setFormData] = useState<{
```
laurenzlong, line 36: this should be defined as a Type, rather than inline like this.

PR #1073, opened 2025-05-07
feat: support headless advanced write 

dionlow:
### Summary
This PR adds support for updating the write features in the config. 
- support enable /disable write per objectName
- support value defaults
- support field settings

use `_default` for value defaults in field settings

#### testing
code is in headless folder non public, to be tested in sample app


jatinsandilya (approved):
looks good!

PR #1091, opened 2025-05-09
fix: remove oauthwindow actions to close window

dionlow:
### Summary 
removes check for oauthWindow. 
we aren't guaranteed access to the window since the window will be coming from a different origin (hubspot). 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Opener-Policy

- we should consider redirecting to the clients website

### DIFF
--- a/src/components/auth/Oauth/AuthorizationCode/OAuthWindow/OAuthWindow.tsx
+++ b/src/components/auth/Oauth/AuthorizationCode/OAuthWindow/OAuthWindow.tsx
@@ -82,7 +82,7 @@ export function OAuthWindow({
       console.error(
         "Window closed and no connection was found. Please try again.",
       );
-      onError?.("Window closed prematurely. Please wait and try again.");
+      // onError?.("Window closed prematurely. Please wait and try again.");
     } else if (connectionId) {
       onError?.(null);
     }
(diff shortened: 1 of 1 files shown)


src/components/auth/Oauth/AuthorizationCode/OAuthWindow/windowHelpers.tsx
```diff
@@ -67,7 +67,7 @@ export function useReceiveMessageEventHandler(
             },
             () => [connection],
           );
-          oauthWindow?.close(); // only close the window if connection is successful
```
laurenzlong, line 70: as discussed, don't remove

laurenzlong (approved):

src/components/auth/Oauth/AuthorizationCode/OAuthWindow/windowHelpers.tsx
```diff
@@ -88,6 +88,6 @@ export function useReceiveMessageEventHandler(
         // do not close the window if error occurs
       }
     },
-    [oauthWindow, onError, queryClient, setConnectionId, onSuccessConnect],
+    [onError, queryClient, setConnectionId, onSuccessConnect],
```
laurenzlong, line 91: revert this line too

PR #1096, opened 2025-05-12
refactor: noWorkspaceOauthFlow2

dionlow:
### Summary
attempts to build the oauth logic from scratch in attempt to debug oauth flow

- Uses own window logic. Creates one window listener on the parent app and only removes when unmounted (no fancy intervals)
- Switches from cached url fetching to async url trigger -> less use states to track. No additional callbacks to track.
- self-contained only to NoWorkspaceOauth flow; will move forward with Workspace fix correct.
- less error handling logic
- refetches connections before submitting for new url

### notes
If this code doesn't pass in production, the likely issue is that we need to follow the standard where the client origin is passed in the callback for message broadcast. Currently now set to '*' in the backend; the message should be broadcasted to any receiver in the browser.

### demo

hubspot-oauth

### DIFF
--- a/src/components/auth/Oauth/AuthorizationCode/NoWorkspaceOauthFlow2/NoWorkspaceOauthFlow2.tsx
+++ b/src/components/auth/Oauth/AuthorizationCode/NoWorkspaceOauthFlow2/NoWorkspaceOauthFlow2.tsx
@@ -1,0 +1,145 @@
+/**
+ * OAuth flow for any providers that do not require the consumer to enter a workspace first.
+ */
+
+import { useEffect, useRef, useState } from "react";
+import { useQueryClient } from "@tanstack/react-query";
+import { useProject } from "src/context/ProjectContextProvider";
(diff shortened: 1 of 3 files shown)


src/components/auth/Oauth/AuthorizationCode/NoWorkspaceOauthFlow2/NoWorkspaceOauthFlow2.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+    try {
+      const { data: connections } = await refetchConnections();
+      if (connections && connections.length > 0) {
+        // first refetch connection before attempting to re-auth
+        console.debug("connections found");
+        return;
+      }
+    } catch {
+      console.debug("no connections found");
```
laurenzlong, line 99: error with fetching connections

laurenzlong (approved):

PR #1102, opened 2025-05-20
[ENG-2258] feat: add update auth flow to hydrated revision

dionlow:
### Summary 
problem: when the user tries to connect via InstallProvider, the component may error out because the token needs to be refreshed. The ability to update the authentication does not show up since the revision is not available.

solution: the hydrated revision error screen adds the ability to re-authenticate as well as delete connection. 

#### Screenshot
Screenshot 2025-05-20 at 2 13 14 PM


RajatPawar (approved):
My bad, I missed this. Approved!

PR #1110, opened 2025-05-23
feat: add callbacks and docs to CRUD installation hook

dionlow:
### Summary 
adds callbacks so that the builder has access to the success, error, and settled state given by tanstack-query. 

### notes:
headless folder is not exposed yet.

### DIFF
--- a/src/headless/installation/useCreateInstallation.ts
+++ b/src/headless/installation/useCreateInstallation.ts
@@ -1,6 +1,7 @@
 import {
   ConfigContent,
   CreateInstallationOperationRequest,
+  Installation,
 } from "@generated/api/src";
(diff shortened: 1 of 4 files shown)


src/headless/installation/useCreateInstallation.ts
```diff
@@ -30,7 +31,17 @@ export function useCreateInstallation() {
     errorMsg,
   } = useCreateInstallationMutation();
 
-  const createInstallation = (config: ConfigContent) => {
+  const createInstallation = ({
```
dionlow, line 34: switch to brackets (object) to be less error prone

laurenzlong (approved):

PR #1124, opened 2025-05-27
[2313] refactor: rename sync to reset

dionlow:
### Summary 
customer feedback said sync was a little confusing. renames sync to reset. 

- existing reset call is unused in demos. (we don't support passing in initial config yet)
- switched implementation to only use reset

### DIFF
--- a/src/headless/config/useConfigHelper.tsx
+++ b/src/headless/config/useConfigHelper.tsx
@@ -59,17 +59,12 @@ export function useConfigHelper(
 ) {
   const [draft, setDraft] =
     useState<UpdateInstallationConfigContent>(initialConfig);
-  const [initial] = useState(initialConfig); // For reset
 
   const { getReadObject: getReadObjectFromManifest, data: manifest } =
     useManifest();
   const { installation } = useInstallation();
 
   const get = useCallback(() => draft, [draft]);
-
-  const reset = useCallback(() => {
-    setDraft(initial);
-  }, [initial]);
 
   /**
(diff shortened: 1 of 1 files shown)


src/headless/config/useConfigHelper.tsx
```diff
@@ -270,32 +265,26 @@ export function useConfigHelper(
     [draft.write?.objects],
   );
 
-  // add loading state
-  const [isSyncing, setIsSyncing] = useState(false);
```
dionlow, line 274: no promise, so no local state needed.

src/headless/config/useConfigHelper.tsx
```diff
@@ -270,32 +265,26 @@ export function useConfigHelper(
...
     // sync the installation config to the local config
-    syncInstallationConfig();
-  }, [installation, syncInstallationConfig]);
+    reset();
+  }, [installation, reset]);
 
   return {
     draft,
     get,
-    syncInstallationConfig,
```
laurenzlong, line 295: I'm a bit confused, the code doesn't seem to reflect the title/description. Seems like this PR is just removing `syncInstallationConfig` and builders should directly use `reset`?

jlimatampersand (approved):

src/headless/config/useConfigHelper.tsx
dionlow: yeah the implementation of reset is unclear; there was potential to set the config manually if the builder would like to pass in the initial config; but resetting to the local config makes way more sense.

dionlow:
update description

PR #1157, opened 2025-06-05
feat: add paginated checkboxes to optional fields

dionlow:


src/components/Configure/content/fields/OptionalFields/OptionalFieldsV2.tsx
```diff
@@ -59,35 +71,16 @@ export function OptionalFieldsV2() {
...
-                  />
-                );
-              }
-              return null;
-            })}
-          </CheckboxFieldsContainer>
-        </CheckboxGroup>
+        <CheckboxPagination
+          // set key so component is re-rendered when optional fields are updated
+          key={`${selectedObjectName}-${readOptionalFields?.length}`}
```
dionlow, line 76: without this key, the page will be saved across different objects causing some pages that don't exist (page 9 of 23). the key will force the page to re-render when the object changes

laurenzlong (approved):
Looks good! Minor tweak: can you try make the "1 of 2" text gray instead of black and see how it looks?

dionlow:
Screenshot 2025-06-09 at 2 26 03 PM

 secondary color typography.

PR #1217, opened 2025-07-18
feat: add jwt token provider to ampersand context

dionlow:
### Summary
This PR adds the ability for the builder to give a callback `getToken` to get a JWT. Our API service will smart detect either an API key or a `getToken` option and use if only one is provided.
- adds `jose` library to handle jwt token expiration / decoding
- adds jwt token provider to ampersand context
- refactor api service to use jwt token provider 
- refactors api service to use API key and JWT headers. 

#### JWT Logic
1. `getToken` gets token from builder's backend
2. `amp-labs/react` api service will look in localstorage for a cached jwt token or call getToken to use / set in cache
3. cached jwt will be loaded into JWT provider to be used in the API service
4. expired JWT will be removed from local storage upon mount

### DIFF
--- a/src/context/AmpersandContextProvider/README.md
+++ b/src/context/AmpersandContextProvider/README.md
@@ -1,0 +1,93 @@
+# AmpersandProvider with JWT Token Support
+
+The `AmpersandProvider` now supports both API key and JWT token authentication methods.
+
+## Usage
+
+### API Key Authentication (Existing)
+
+```tsx
+import { AmpersandProvider } from './AmpersandContextProvider';
+
+function App() {
+  return (
+    <AmpersandProvider
+      options={{
+        apiKey: "your-api-key-here",
+        project: "your-project-id"
+      }}
+    >
+      {/* Your app components */}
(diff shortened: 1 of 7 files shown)


src/context/AmpersandContextProvider/README.md
```diff
@@ -0,0 +1,93 @@
+# AmpersandProvider with JWT Token Support
```
dionlow, line 1: README auto generated - might remove.

src/services/api.ts
```diff
@@ -110,6 +111,35 @@ export const setApi = (api: ApiService) => {
...
 
+// Authentication helper functions
+const createApiKeyAuth = (apiKey: string) => ({
+  header: "X-Api-Key",
+  value: apiKey,
+});
+
+const createJwtAuth = async (jwtToken: any) => {
+  try {
+    const token = await jwtToken.getToken("default", "default");
```
dionlow, line 122: need to fix

dionlow:
decided to use session storage instead of local storage

PR #1219, opened 2025-07-24
[ENG-2416] feat: add jwt provider session storage part 1

dionlow:
### Summary
This PR adds the ability for the builder to give a callback `getToken` to get a JWT. The API service will detect either an API key or a `getToken` option and use if only one is provided.
- adds `jose` library to handle jwt token expiration / decoding
- adds jwt token provider to ampersand context
- refactor api service to use jwt token provider 
- refactors api service to use API key and JWT headers. 
- `ENABLE_JWT_AUTH_FF` hard coded feature flag blocks usage of JWT for time being though the param is available in AmpersandProvider

#### Supported
- `InstallationProvider` (supported)
- `ConnectProvider` (supported)
- `InstallIntegration` (not supported yet)

#### Tested 
test API key workflow for ConnectProvider

test jwt flow
1. Turn on feature flag `ENABLE_JWT_AUTH_FF=true`
2. In react `run yarn build`
3. In mailmonkey add a getToken callback

```
const getToken = async (consumerRef: string, groupRef: string) => {
  return await Promise.resolve("your-jwt-token");
}

// add getToken to AmpersandProvider 
 <AmpersandProvider options={{
  // apiKey: defaultApiKey, // comment out apiKey; cannot have both
     projectId: defaultProjectId,
    getToken: getToken
}}>
```
4. Set up preview environment (`source vars.sh` & `./run_mailmonkey.sh` )
5. Generate JWT token using `go run generateJWT.go` in `feat/jwt-integration`
6. Copy in JWT token to mailmonkey
7. test `localhost:3000/connect` (`ConnectProvider`)

#### JWT Logic
1. `getToken` gets token from builder's backend
2. `amp-labs/react` api service will look in local cache (react) and `session storage`(browser) for a cached jwt token or call `getToken` to use and set in cache/session storage.
3. expired JWT will be removed from storage upon mount and when getting a cached token that has expired
4. cached jwt will be loaded into JWT provider to be used in the API service

##### Session Storage vs Local Storage
Session Storage: same session / same tab - auto clear when tab closes
Local Storage: works across tabs, need manual deletion

Session storage is marginally safer as the need to clear cache is a little less important considering that user will likely close tabs. Session storage will still allow the app to work when user refreshes the page. 

##### Not included
To be most secure, we'll want to implement a http cookie, but 10 min jwt tokens will stil minimize access.

##### README pr
https://github.com/amp-labs/react/pull/1220

### DIFF
--- a/src/context/ApiKeyContextProvider.tsx
+++ b/src/context/ApiKeyContextProvider.tsx
@@ -1,20 +1,10 @@
 import { createContext, useContext } from "react";
 
-import { useJwtToken } from "./JwtTokenContextProvider";
-
 export const ApiKeyContext = createContext<string | null>(null);
 
 export const ApiKeyProvider = ApiKeyContext.Provider;
 
 export const useApiKey = () => {
   const apiKey = useContext(ApiKeyContext);
-  const { getToken } = useJwtToken();
-
-  if (apiKey === null && getToken == null) {
-    console.error(
-      "useApiKey must be used within an ApiKeyProvider, or there is no JWT token callback",
-    );
-  }
-
   return apiKey;
 };
--- a/src/context/JwtTokenContextProvider.tsx
+++ b/src/context/JwtTokenContextProvider.tsx
@@ -7,24 +7,20 @@ import {
 } from "react";
 import { jwtVerify } from "jose";
 
-// Token cache types
 interface TokenCacheEntry {
   token: string;
   expiresAt: number;
 }
 
-/**
- * Create a cache key for the JWT token
- * @param consumerRef - The consumer reference
- * @param groupRef - The group reference
- * @returns The cache key
- */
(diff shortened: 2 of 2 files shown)


src/services/api.ts
```diff
@@ -110,6 +112,37 @@ export const setApi = (api: ApiService) => {
...
+    basePath: AMP_API_ROOT,
+    headers: {
+      "X-Amp-Client": "react",
+      "X-Amp-Client-Version": LIB_VERSION,
+      [authHeader]: authValue,
+    },
+  });
+
+// TODO: remove this flag when we have a proper JWT auth flow
+const ENABLE_JWT_AUTH_FF = false;
```
dionlow, line 144: turn this to true to test; we can also inject a env flag here later.

PR #1246, opened 2025-08-12
[ENG-2417] build: add installation provider to install integration

dionlow:
### Summary
adding installation provider gives access to groupRef and consumerRef to the apiService in InstallIntegration -- needed for JWT support.
- adds InstallationProvider to InstallIntegration
- JWT support is still disabled in InstallIntegration

### Steps
1. ~~support JWT through InstallationProvider~~
2. **add InstallationProvider to InstallIntegration**
3. refactor legacy setAndCreate calls to use react-query / centralized api service
4. test InstallIntegration Create and Update flow using JWT
5. remove error dialog for InstallIntegration JWT support


src/components/Configure/InstallIntegration.tsx
```diff
@@ -180,4 +181,39 @@ export function InstallIntegration({
       </InstallIntegrationProvider>
     </div>
   );
+};
+
+export function InstallIntegration({
```
dionlow, line 186: We surround the entire existing InstallIntegration component with the InstallationProvider which is now need for all api calls including the projectQuery and listIntegrationsQuery at the top level.

dionlow: no visual changes in this PR.

PR #1248, opened 2025-08-12
[ENG-2476] feat: checkbox search

dionlow:
### Summary
checkbox components are paginated, but with hundreds of fields a searching ux is need. This PR adds the search functionality to the paginated checkbox component and refactors optional fields and write to use the new search ux.
- pagination moved down since less likely to be used with search
- filter applies to groups of fields 
- note the buttons moving to the bottom still makes the ux clucky to navigate to next page since the page length changes. should be less of a concern with search being instructed by builders to find specific fields.

#### demo
search-checkbox-ux

### DIFF
--- a/src/components/ui-base/Checkbox/CheckboxPagination.tsx
+++ b/src/components/ui-base/Checkbox/CheckboxPagination.tsx
@@ -93,7 +93,7 @@ export function CheckboxPagination({
         <SelectAllCheckbox
           id="select-all-fields"
           isChecked={areAllFilteredItemsChecked}
-          label={`Select all ${filteredItems.length} fields`}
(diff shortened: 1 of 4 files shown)


src/components/ui-base/Checkbox/CheckboxPagination.tsx
```diff
@@ -20,32 +21,100 @@ export interface CheckboxItem {
...
+            value={searchTerm}
+            onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+              handleSearchChange(e.target.value)
+            }
+            className={styles.searchInput}
+          />
+        </div>
+      )}
+
+      {showSelectAll && items.length >= 2 && (
```
jlimatampersand, line 92: I think it would be nice to have a comment on what this logic means.

jlimatampersand (approved):

PR #1340, opened 2025-09-20
fix/rm/feat: simplify value mapping

dionlow:
### Summary  
Tracking the `isModified` state has resulted in unnecessary complexity as the product has scaled. With multiple states updating a legacy draft state and a legacy saved state, the “source of truth” has been ambiguous, leading to bugs that are difficult to debug. A state rewrite was needed to debug this issue. 

This PR simplifies the `ValueMappings` implementation in the following ways:  
- **Remove inline validation**: Since we now enforce uniqueness at the selection step (via filter), post-input validation is no longer required. This reduces maintenance burden.  
- **Remove `isValueMappingsModified` state tracking**: No longer stored in state; simplifies logic and reduces potential divergence.  
- **Add `isValueMappingsEqual` helper**: Ensures `undefined` (server) and `{}` (local) are treated as equal (not yet modified).  

### Other details
cleaned up value mappings  with utility functions
- getFieldDisplayName
- validateFieldMapping
- getAvailableOptions

---

### Legacy Data Rendering (pre-reactQuery)  
1. Server pulls data → reset saved value mappings (from config)  
2. Local draft created  
3. Local draft modified  
4. Trigger `isEqual` check  
5. Update `isValueMappingsModified`  
6. Submit  
7. Update server saved  

This approach was fragile—errors were hard to locate, and debugging required juggling multiple configs (`valueMappings`, `readConfig`, `writeConfig`). Tracking the "Save" button state became overly complex.  

Debugging saw a problem where after create installation was fired off, the submit button was still enabled. This rendering issue was caused because the savedValueMapping (server) state was not updated locally calling a manual reset. Instead of navigating how to call a reset; we removed the tracking; and implemented a new data flow architecture below.

---

### New Approach (with React Query)  
1. `useQuery` pulls installation → **server config is the single source of truth**  
2. `selectedValueMappings` tracks local draft only  
3. `isValueMappingsModified` becomes derived state (compare 1 & 2 directly)  

---

### Takeaways  
- The MVP blurred the separation between server state and local draft state.  
- That design was later scaffolded and “patched” (including by AI tools), but the flawed foundation made debugging impossible.  
- Removing explicit local tracking of `isModified` eliminates a recurring class of bugs.  
- **Recommendation**: Apply this pattern consistently across the form—remove local state management of modified state wherever possible.  (starting with isWriteModified, isReadModified, isOptional etc...)

### Testing
simplify-rendering-state

### DIFF
--- a/src/components/Configure/content/fields/ValueMapping/ValueMappingItem.tsx
+++ b/src/components/Configure/content/fields/ValueMapping/ValueMappingItem.tsx
@@ -1,7 +1,6 @@
 import { useCallback, useEffect, useMemo, useState } from "react";
 import { Button } from "src/components/ui-base/Button";
 import { ComboBox } from "src/components/ui-base/ComboBox/ComboBox";
-import { ErrorBoundary, useErrorState } from "src/context/ErrorContextProvider";
 
 import { useSelectedConfigureState } from "../../useSelectedConfigureState";
 
@@ -22,7 +21,6 @@ interface ValueMappingItemProps {
   fieldName: string;
   onSelectChange: (e: React.ChangeEvent<HTMLSelectElement>) => void;
   allValueOptions: ValueOption[];
-  hasError?: boolean;
 }
 
 export function ValueMappingItem({
@@ -30,14 +28,10 @@ export function ValueMappingItem({
   onSelectChange,
   fieldName,
   allValueOptions,
-  hasError,
 }: ValueMappingItemProps) {
   const { configureState, selectedObjectName, setConfigureState } =
     useSelectedConfigureState();
   const [disabled, setDisabled] = useState(true);
-
-  const { getError, setError, resetBoundary, isError, removeError } =
-    useErrorState();
 
   const selectedValueMappingForField = useMemo(
     () => configureState?.read?.selectedValueMappings?.[fieldName] || {},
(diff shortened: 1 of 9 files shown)


src/components/Configure/state/ConfigurationStateProvider.tsx
```diff
@@ -133,7 +133,6 @@ export function ConfigurationProvider({
           if (readDraft) {
             readDraft.isOptionalFieldsModified = false;
             readDraft.isRequiredMapFieldsModified = false;
-            readDraft.isValueMappingsModified = false;
```
dionlow, line 136: removes local state tracking

src/components/Configure/state/utils.ts
```diff
@@ -102,11 +102,9 @@ const generateConfigurationStateRead = (
     selectedValueMappings,
     isOptionalFieldsModified: false,
     isRequiredMapFieldsModified: false,
-    isValueMappingsModified: false,
     savedConfig: {
       optionalFields: optionalFieldsSaved, // from config
       requiredMapFields: requiredMapFieldsSaved, // from config
-      selectedValueMappings,
```
dionlow, line 109: removes saved config; use serverConfig directly (react-query handles async state)

src/components/Configure/content/fields/ValueMapping/ValueMappingItem.tsx
```diff
@@ -73,37 +56,6 @@ export function ValueMappingItem({
     (item: { value: string } | null) => {
       if (!item) return;
 
-      // Validate that target values are unique within the same field
```
dionlow, line 76: removes post input validation; the new ux gaurentees unique values so this code is not needed.

eberle1080 (approved):

PR #1347, opened 2025-09-23
[ENG-2596] refactor: remove write modified state

dionlow:
### Summary
Continued work to migrate write modified state to derived state to reduce state management complexity.
- fix: headless write objects update mask
- add write objects to update mask if it is modified from server state

testWriteObjects

### DIFF
--- a/src/components/Configure/state/utils.ts
+++ b/src/components/Configure/state/utils.ts
@@ -1,4 +1,3 @@
-import isEqual from "lodash.isequal";
 import {
   Config,
   HydratedIntegrationFieldExistent,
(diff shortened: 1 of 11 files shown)


src/components/Configure/state/utils.ts
```diff
@@ -27,12 +26,30 @@ import {
   isIntegrationFieldMapping,
 } from "../utils";
 
-// uses lodash deep equality check to compare two saved write objects (typed checked)
+// compares two write objects by checking if they have the same objectNames
+// implementation detail: sdk generates selectedValueDefaults, selectedFieldSettings;
+// todo: switch to not pass the undefined values to the sdk
```
laurenzlong, line 31: this sentence is confusing, I think what you mean is "update openapi spec so the generated SDK no longer have undefined values for selectedValueDefaults, selectedFieldSettings"

nothing is passing undefined values to the SDK, the SDK is generating fields with undefined values

laurenzlong (approved):

PR #1353, opened 2025-09-30
[ENG-2411] feat: track dynamic mappings in the config

RajatPawar:
Previously, we only stored selected dynamic mappings in the config. This was an issue because when removing unknown fields from a write request on the server, it could have also been an unmapped dynamic mapping. 

To get around this, we needed to store all dynamic mappings in the config.

## Testing
Screenshot 2025-10-01 at 1 46 46 PM

Screenshot 2025-10-01 at 1 28 01 PM

Screenshot 2025-10-01 at 1 28 19 PM
Screenshot 2025-10-01 at 1 42 07 PM

Screenshot 2025-10-01 at 1 43 53 PM

## Changes
- Defines dynamic mapping types in https://github.com/amp-labs/openapi/pull/225/files to store all dynamic mappings that were available for the installation
- Removes the old dynamic mapping types that were manually defined in favor of generated types
- Sends over ALL dynamic mappings in an object's config, giving us a complete picture


src/components/Configure/state/utils.ts
```diff
@@ -231,6 +233,25 @@ export const generateSelectedFieldMappingsFromConfigureState = (
   return selectedRequiredMapFieldsSubmit;
 };
 
+/**
+ * generates dynamicFieldMappings object for Config from fieldMapping prop
+ * @param objectName - object name to get field mappings for
+ * @param fieldMapping - dynamic field mappings from InstallIntegration component
+ * @returns
+ */
+export const getObjectDynamicMappings = (
```
RajatPawar, line 242: simple function to get the object's dynamic mappings from the entire object

src/components/Configure/InstallIntegration.tsx
```diff
@@ -20,40 +21,11 @@ import { HydratedRevisionProvider } from "./state/HydratedRevisionContext";
...
-   * The name of the field in SaaS provider, if present, then we will not prompt the user to map the field.
-   */
-  fieldName?: string;
-};
-
 /**
  * A map of object names to FieldMappingEntry arrays, with each FieldMappingEntry representing a field.
  */
 export type FieldMapping = {
-  [key: string]: Array<FieldMappingEntry>;
```
dionlow, line 56: Depending on implementation; this could be a breaking change if FieldMappingEntry was used in the parent App. noting for version number.

src/components/Configure/InstallIntegration.tsx
```diff
@@ -20,40 +21,11 @@ import { HydratedRevisionProvider } from "./state/HydratedRevisionContext";
...
-   */
-  mappedValues?: MappedValue[];
-  /**
-   * The name of the field in SaaS provider, if present, then we will not prompt the user to map the field.
-   */
-  fieldName?: string;
-};
-
 /**
  * A map of object names to FieldMappingEntry arrays, with each FieldMappingEntry representing a field.
```
dionlow, line 25: Type in comment should be updated.

dionlow:
Just to clarify: we're storing a copy of the dynamic mappings array being passed in so the server has context? 

Would it make sense to separate the input entirely so we aren't coupling the mappings together? Would builders still pass in a dynamic mappings array in headless?

src/components/Configure/InstallIntegration.tsx
RajatPawar: I used the example from our docs, and added it under a testing section in our PR desc

dionlow (approved):
lgtm

PR #1646, opened 2026-09-03
[CON-3514] adding region parameter

buildstuff-2026:
Adds an internal `region` option to `AmpersandProvider`. With `region: "eu"`, all API requests go to `https://api.eu.withampersand.com` instead of the default `https://api.withampersand.com`.

  ```ts
  const options = { project: 'PROJECT', apiKey: 'API_KEY', region: 'eu' };
  <AmpersandProvider options={options}>…</AmpersandProvider>
```

#### Before the change the default endpoint was US
Screenshot 2026-09-03 at 13 03 07

#### After the change when passing EU it uses eu endpoint
Screenshot 2026-09-03 at 13 19 39

####  when someone sets a eu region on the lib, but the  `REACT_APP_AMP_SERVER` takes precedence
Screenshot 2026-09-04 at 14 17 00


RajatPawar (approved):
LGTM, can you also add a test screen of when someone sets a `eu` region on the lib, but the REACT_AMP_API_URL takes precedence? (i might be misremembering the exact env name)

